PR #55, opened 2025-04-02
Fix equity link breaking for Gumroad workers

MayaRainer:
Currently, the cap table is hardcoded to allow Antiwork workers to view it. This is specific to hourly workers however, and is now breaking for project-based workers. Since this is probably now-obsolete legacy code as anyone working at Antiwork is also an investor, this PR just removes it.

### DIFF
--- a/apps/next/trpc/routes/capTable.ts
+++ b/apps/next/trpc/routes/capTable.ts
@@ -3,7 +3,6 @@ import { and, desc, eq, inArray, or, sql, sum } from "drizzle-orm";
 import { omit, pick } from "lodash-es";
 import { z } from "zod";
 import { db } from "@/db";
-import { PayRateType } from "@/db/enums";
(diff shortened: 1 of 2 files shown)


apps/next/trpc/routes/capTable.ts
```diff
@@ -20,12 +19,7 @@ import { companyProcedure, createRouter } from "@/trpc";
 export const capTableRouter = createRouter({
   show: companyProcedure.input(z.object({ newSchema: z.boolean().optional() })).query(async ({ ctx, input }) => {
     const isAdminOrLawyer = !!(ctx.companyAdministrator || ctx.companyLawyer);
-    const isWorkerForGumroad =
```
MayaRainer, line 23: @slavingia lmk if there's a reason you'd like to keep this!

Issue #18, opened 2025-03-27
Shadcn migration


ershad:
**Goal:**

To improve UI consistency, maintainability, and streamline development, we need to migrate our custom UI components to the defaults provided by Shadcn/ui. This involves replacing components currently located in `apps/next/components` with their Shadcn equivalents, which will reside in `apps/next/components/ui`.

**Current State:**

Our codebase contains a mix of:

1.  **Custom components** located directly in `apps/next/components`. These need to be replaced.
2.  **Components located in `apps/next/components/ui`**. These might already be Shadcn components or implementations closely resembling them. These should be verified against Shadcn standards and updated/replaced if necessary for full consistency.

**Migration Plan:**

1. Use shadcn@canary CLI to add components
2. Migrate all usages of existing components to the newly created ones
3. Delete the old component files after successful migration
4. Verify functionality after each component migration

*   [ ] `Input` (50+ usages)
*   [ ] `Select` (50+ combined usages with `/ui/select`) - Our Select; Switch to using Combobox instead of Shadcn's Select
*   [ ] `Combobox`  
*   [ ] `Form` (50+ usages - uses `FormSection`)
*   [ ] `Label` (~17+ usages)
*   [ ] `Select`

*   [ ] Modal
*   [ ] `Calendar / Date Picker`

Easiest to get rid of Input, FormSection, Select, Label, and the old way of doing forms/validation all at once than trying to migrate fifty usages at once, and avoids conflicts between them.

hellosatyajit:
Hey @ershad I'd love to take this up.

slavingia:
We don't assign issues to devs. You can just open a PR or ask specific questions for clarity.

hellosatyajit:
understood, thanks

adithyaakrishna:
@ershad @slavingia Does this PR only aim to add new components or does it include removing of old and then updating all files to use the new shadcn components?

slavingia:
The latter

adithyaakrishna:
Perfect, I thought soo too. Im in the process of it, but the docs doesnt help much to setup the project locally :/

slavingia:
To confirm the basics, did you run this?

![Image](https://github.com/user-attachments/assets/0b464d49-4c81-45cd-9123-76b9c527755f)

noobCode-69:
hi @slavingia I'm thinking of contributing but creating envs would take years 😔

MayaRainer:
[...]

PR #57, opened 2025-04-03
feat(alert): created alert component using shadcn library

somyanath:
This PR completes the following component from #18  using shadcn library

- [x] alert


MayaRainer:
@somyanath _a When migrating a component to Shadcn, please migrate all usages of existing components (in this case `Notice`) as well.

PR #58, opened 2025-04-03
feat(roles): Make pay rate optional for project-based roles

### LINKED ISSUE #8: Make pay rate optional for project-based roles
Would be nice to make this optional generally. It's quite confusing to onboard a vendor with a default project-based rate. Some may be $10,000 a month also, not just "project" so I feel like we should just remove that completely.

_Originally posted by @slavingia in https://github.com/antiwork/flexile/pull/4108#discussion_r1989904252_
            

Also allow customizing the "noun" the person is doing, making it optional too. Project can be the default, but it can change to month, article, video edit, etc.

asadath1395:
Fixes #8

**Allowing users to specify sub-type like project/article**

https://github.com/user-attachments/assets/39cc3261-2934-465e-8a41-2c2997911166

**Allowing role creation without any pay rate**

https://github.com/user-attachments/assets/2d4daa7a-f07e-4c5b-9291-1c38cc1d77ba

#### TODO
- Add/Fix tests

### DIFF
--- a/apps/next/db/schema.ts
+++ b/apps/next/db/schema.ts
@@ -267,9 +267,10 @@ export const companyRoleRates = pgTable(
     updatedAt: timestamp("updated_at", { precision: 6, mode: "date" })
       .notNull()
       .$onUpdate(() => new Date()),
-    payRateInSubunits: integer("pay_rate_in_subunits").notNull(),
+    payRateInSubunits: integer("pay_rate_in_subunits"),
     payRateCurrency: varchar("pay_rate_currency").default("usd").notNull(),
     trialPayRateInSubunits: integer("trial_pay_rate_in_subunits").notNull(),
+    payPer: varchar("pay_per").default("project"),
   },
   (table) => [
     index("index_company_role_rates_on_company_role_id").using(
--- a/apps/next/trpc/routes/roles/index.ts
+++ b/apps/next/trpc/routes/roles/index.ts
@@ -23,11 +23,15 @@ const inputSchema = createInsertSchema(companyRoles)
     trialEnabled: true,
   })
   .merge(
-    createInsertSchema(companyRoleRates, { payRateType: z.nativeEnum(PayRateType) }).pick({
-      payRateInSubunits: true,
-      payRateType: true,
-      trialPayRateInSubunits: true,
-    }),
+    createInsertSchema(companyRoleRates, { payRateType: z.nativeEnum(PayRateType) })
+      .pick({
+        payRateType: true,
+        trialPayRateInSubunits: true,
+      })
+      .extend({
+        payRateInSubunits: z.number().nullable(),
+        payPer: z.string().nullable(),
+      }),
   );
 
 export const rolesRouter = createRouter({
@@ -75,7 +79,7 @@ export const rolesRouter = createRouter({
           "applicationCount",
           "expenseCardsCount",
         ),
-        ...pick(rate, "payRateType", "payRateInSubunits", "trialPayRateInSubunits"),
+        ...pick(rate, "payRateType", "payRateInSubunits", "trialPayRateInSubunits", "payPer"),
       };
     });
   }),
@@ -86,7 +90,7 @@ export const rolesRouter = createRouter({
     const [role] = await db
       .select({
         ...pick(companyRoles, "id", "name", "jobDescription", "activelyHiring", "capitalizedExpense"),
-        ...pick(companyRoleRates, "payRateType", "payRateInSubunits", "trialPayRateInSubunits"),
+        ...pick(companyRoleRates, "payRateType", "payRateInSubunits", "trialPayRateInSubunits", "payPer"),
       })
       .from(companyRoles)
       .innerJoin(companyRoleRates, eq(companyRoles.id, companyRoleRates.companyRoleId))
@@ -124,7 +128,9 @@ export const rolesRouter = createRouter({
       const role = assertDefined(result[0]);
       await tx.insert(companyRoleRates).values({
         companyRoleId: role.id,
-        ...pick(input, "payRateType", "payRateInSubunits", "trialPayRateInSubunits"),
+        ...pick(input, "payRateType", "trialPayRateInSubunits", "payPer"),
+        payRateInSubunits:
+          input.payRateType === PayRateType.ProjectBased ? (input.payRateInSubunits ?? null) : input.payRateInSubunits,
         payRateCurrency: "usd",
       });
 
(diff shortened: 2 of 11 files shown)


asadath1395:
@slavingia Can you take a look at this PR and let me know if i missed something. Thanks

slavingia:
Will cover this Friday as part of the sync

slavingia:
It may good to write a test for this and make sure inviting a project-based contractor works end-to-end for a company, including both sides signing the contract? cc @MayaRainer 

My product feedback is if they select "Custom" which I like more than project-based, there should be a shadcn toggle-switch that says "Specify a default amount" which upon turning on, would show the other two fields, first:

Amount

then

Unit of work (which would fill in the suffix above)

Here's a v0 of that in action: https://v0.dev/chat/custom-feedback-toggle-jUnB5Qe4Zdr?b=b_IFh0kI5wICX

slavingia:
Its more like "contractor can bill whatever they want" and if a rate is chosen, its just the default amount that's filled in

Customizablity should exist for those roles when filling out an invoice e.g they could have some line items at different hourly rates.

asadath1395:
@slavingia Any reason for closing this PR?

slavingia:
Wasn't ready to merge. Reopen once it is

apps/rails/db/schema.rb
```diff
@@ -290,9 +290,11 @@
     t.bigint "company_role_id", null: false
     t.datetime "created_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
     t.datetime "updated_at", null: false
-    t.integer "pay_rate_in_subunits", null: false
+    t.integer "pay_rate_in_subunits"
     t.string "pay_rate_currency", default: "usd", null: false
-    t.integer "trial_pay_rate_in_subunits", null: false
+    t.integer "trial_pay_rate_in_subunits"
```
raulpopadineti, line 295: Nullifying these has bigger implications since it's used in applications, contract templates, invoices and this will break a lot of functionality. Please make sure to test inviting a contractor with a custom role that doesn't have a default amount specified.

Rails model validations need to be updated too.

apps/rails/db/schema.rb
```diff
@@ -290,9 +290,11 @@
...
+    t.string "pay_per", default: "project"
```
raulpopadineti, line 296: This must've been added by mistake. Please remove it.

e2e/tests/company/administrator/new-contract.spec.ts
```diff
@@ -102,7 +102,7 @@ test.describe("New Contractor", () => {
       __role: "Hourly Role 1",
       __targetAnnualHours: "Target Annual Hours: 1,100",
       __maximumFee:
-        'Maximum fee payable to Contractor on this Project Assignment, including all items in the first two paragraphs above is $152,460 (the "Maximum Fee").',
+        'Maximum fee payable to Contractor on this Project Assignment, including all items in the first two paragraphs above is $1,52,460 (the "Maximum Fee").',
```
raulpopadineti, line 105: Please revert this as it's incorrect.

apps/next/trpc/routes/roles/index.ts
```diff
@@ -124,7 +128,9 @@ export const rolesRouter = createRouter({
       const role = assertDefined(result[0]);
       await tx.insert(companyRoleRates).values({
         companyRoleId: role.id,
-        ...pick(input, "payRateType", "payRateInSubunits", "trialPayRateInSubunits"),
+        ...pick(input, "payRateType", "trialPayRateInSubunits", "unitOfWork"),
+        payRateInSubunits:
+          input.payRateType === PayRateType.ProjectBased ? (input.payRateInSubunits ?? null) : input.payRateInSubunits,
```
raulpopadineti, line 133: Why is this change necessary?

apps/rails/db/migrate/20250405080045_add_unit_of_work_to_company_role_rates.rb
```diff
@@ -0,0 +1,6 @@
+class AddUnitOfWorkToCompanyRoleRates < ActiveRecord::Migration[7.1]
+  def change
+    add_column :company_role_rates, :unit_of_work, :string, default: "project"
+    change_column_null :company_role_rates, :trial_pay_rate_in_subunits, true
+  end
+end
```
raulpopadineti, line 6: Please split the `change_column_null` into a separate migration. The migration name doesn't reflect this change and it's better to have them separate so it's easier to understand at a single glance what the migration does.

raulpopadineti: It's also missing the `change_column_null` for `pay_rate_in_subunits` column.

apps/next/trpc/routes/roles/index.ts
```diff
@@ -160,7 +166,11 @@ export const rolesRouter = createRouter({
         .update(companyRoleRates)
         .set({
           companyRoleId: role.id,
-          ...pick(input, "payRateType", "payRateInSubunits", "trialPayRateInSubunits"),
+          ...pick(input, "payRateType", "trialPayRateInSubunits", "unitOfWork"),
+          payRateInSubunits:
+            input.payRateType === PayRateType.ProjectBased
+              ? (input.payRateInSubunits ?? null)
+              : input.payRateInSubunits,
```
raulpopadineti, line 173: Same question here as above ^

apps/next/db/schema.ts
```diff
@@ -267,9 +267,10 @@ export const companyRoleRates = pgTable(
     updatedAt: timestamp("updated_at", { precision: 6, mode: "date" })
       .notNull()
       .$onUpdate(() => new Date()),
-    payRateInSubunits: integer("pay_rate_in_subunits").notNull(),
+    payRateInSubunits: integer("pay_rate_in_subunits"),
     payRateCurrency: varchar("pay_rate_currency").default("usd").notNull(),
     trialPayRateInSubunits: integer("trial_pay_rate_in_subunits").notNull(),
```
raulpopadineti, line 272: This seems to have been missed.

```suggestion
    trialPayRateInSubunits: integer("trial_pay_rate_in_subunits"),
```

raulpopadineti (changes requested):
@asadath1395 _a there are still parts missing from this feature to be completed and there are quite a handful of inconsistencies when it comes to migrations and dealing with the `null` column values. 

We don't want to default to `0` on the frontend when there's no default amount set for the role, but we want to hide all the sections that are affected by this completely.

Also, for document templates, we should adjust [the auto-filled pay rate field](https://github.com/antiwork/flexile/blob/c902667aca1cdca2a3b8a318d15c3347c5e12500/apps/next/trpc/routes/documents/templates.ts#L144-L150) for this change and add an e2e spec for inviting a new contractor w/ a custom role that has no default amount set to ensure everything's still working as expected.

slavingia:
Since this is getting old, I'm closing this. Can re-open with the migration to make null and the other changes in a new PR.

PR #60, opened 2025-04-03
Remove #pay_rate_type_cannot_change validation

MayaRainer:
Currently, the `CompanyWorker` rails model has a validation to prevent changing the pay rate type. This legacy code is breaking some flows, so we should remove it.


MayaRainer (commented):
@slavingia _a lmk if you can think of a reason to keep this, otherwise this should fix the issue you were having!

slavingia:
We'll learn what issues exist shortly! There may be some issues with equity vesting etc but hard to know without doing an extensive git blame.

slavingia (approved):

PR #74, opened 2025-04-07
Enable Playwright parallelism

### LINKED ISSUE #45: Enable Playwright parallelism
Currently, Playwright tests can only be run serially, which slows them down a lot. To fix this, we need to get rid of `truncateTablesBeforeEachTest`, and remove the hacky patch we applied to the Next testing proxy to be able to test Inngest. Ideally, we can find some other way to integrate Inngest into our E2E tests; otherwise we'll have to test it separately. Quite a few tests will likely break or be flaky without truncation, but if we use entirely separate datasets for each of them and separate Clerk users for each worker (until #44), we should be able to make it work.

MayaRainer:
Closes #45.


e2e/helpers/auth.ts
```diff
@@ -3,14 +3,37 @@ import { type Page } from "@playwright/test";
 import { db } from "@test/db";
 import { eq } from "drizzle-orm";
 import { users } from "@/db/schema";
+import { assertDefined } from "@/utils/assert";
+
+const clerkTestUsers = [
```
MayaRainer, line 8: When Clerk is gone, this can be simplified a lot, and we could even upgrade the CI worker to the 8-core machine to speed this up further. For now, this works.

patches/next.patch
```diff
@@ -11,37 +11,11 @@ index cae9ec8cd4e4e266df599ea6a6d3a16af8ac6feb..2e74e45e848c2a7a7c1661a2e06e1d73
              cache,
              credentials,
              integrity,
-@@ -83,8 +83,9 @@ function buildResponse(proxyResponse) {
```
MayaRainer, line 14: Removed the override we had here to be able to test Inngest as it was breaking parallelism.

patches/playwright.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/lib/common/process.js b/lib/common/process.js
+index 1fde97528c9a87ab61a28df5814869e2486b9394..be635961239962351586470ceb1ff9f29c37757c 100644
+--- a/lib/common/process.js
++++ b/lib/common/process.js
+@@ -133,7 +133,10 @@ async function gracefullyCloseAndExit(forceExit) {
+ }
+ function sendMessageToParent(message) {
+   try {
++    const originalToJSON = BigInt.prototype.toJSON;
```
MayaRainer, line 9: Works around a bug in Playwright - [opened an issue for it there too](https://github.com/microsoft/playwright/issues/35562).

slavingia (approved):

PR #80, opened 2025-04-08
fix: copying wrong env example file

lukicenturi:
- [x] When first time developer runs `bin/setup`, and then he don't take env from vercel, it will try to copy `.env.sample` to `.env`. Which is not exist (I believe this should be .env.example) instead
- [x] Add .idea and .vscode to the .gitignore


MayaRainer:
Thank you!

PR #116, opened 2025-04-12
Allow CI to run on bot commits

MayaRainer:


MayaRainer:
[Doesn't seem to be possible](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#:~:text=Note-,This%20event%20will%20only%20trigger%20a%20workflow%20run%20if%20the%20workflow%20file%20exists%20on%20the%20default%20branch.,-Runs%20your%20workflow%20when%20branch) to test this on a branch that isn't `main`, so going to have to merge this and push follow-ups if this doesn't work.

MayaRainer:
Looks like this isn't a great solution either as GH doesn't display the workflow run as a part of the PR like this. We may have to revert to something like the previous solution and hardcode a list of bots to allow.

PR #119, opened 2025-04-13
refactor(ui): migrate sheet from components to components/ui

rimildeyjsr:
## Description

1. Removed the implementation of sheet component in `components/Sheet.tsx`
2. Replace all implementations of `components/Sheet.tsx` with `components/ui/sheet.tsx`
3. Adjust the implementation of `components/ui/sheet.tsx` so that the component is nearer in implementation to the previous one

Sub-Issue of https://github.com/antiwork/flexile/issues/18

PS: Used mock data to test the components

### Before (invoices admin list)

Screenshot 2025-04-13 at 11 43 19 AM

### After (invoices admin list)

Screenshot 2025-04-13 at 2 48 07 PM

### Before (documents page)

Screenshot 2025-04-13 at 8 42 53 PM

### After (documents page)

Screenshot 2025-04-13 at 8 43 40 PM

### Before (cap table page)

Screenshot 2025-04-13 at 9 13 01 PM

### After (cap table page)

Screenshot 2025-04-13 at 9 10 53 PM


apps/next/app/documents/page.tsx
```diff
@@ -127,12 +127,17 @@ function Documents({ year }: { year: number }) {
...
-            {formatDate(filingDueDateFor1042S)}, and form 1099-DIV on {formatDate(filingDueDateFor1099DIV)}.
-          </p>
+        <Sheet open modal={false}>
+          <SheetContent side="bottom" hideCloseButton modal={false} className="relative w-full">
+            <SheetHeader>
+              <SheetTitle>Upcoming filing dates for 1099-NEC, 1099-DIV, and 1042-S</SheetTitle>
+
+              <SheetDescription>
+                We will submit form 1099-NEC to the IRS on {formatDate(filingDueDateFor1099NEC)}, form 1042-S on{" "}
+                {formatDate(filingDueDateFor1042S)}, and form 1099-DIV on {formatDate(filingDueDateFor1099DIV)}.
```
MayaRainer, line 137: maybe better to have this text outside of `SheetHeader`? wdyt?

apps/next/components/ui/sheet.tsx
```diff
@@ -32,41 +36,63 @@ function SheetOverlay({ className, ...props }: React.ComponentProps<typeof Sheet
   );
 }
 
+interface SheetContentProps extends React.ComponentProps<typeof SheetPrimitive.Content> {
```
MayaRainer, line 39: Shadcn assumes that `Sheet` will always be used as a modal, which makes this a bit awkward - I think instead of the `hideCloseButton` and `modal` props here we should extract out the portal, overlay, and close button from `SheetContent` and export them separately where necessary, and then compose them in usages the same way the `Dialog` component works. Wdyt?

Also think we can get rid of `primary` and just use the default styling here, but lmk if you disagree @slavingia

MayaRainer (commented):
@rimildeyjsr _a Tests are currently failing as the sheet seems to be rendered as a modal over the content breaking interaction with it - could you fix that please? also added two suggestions

apps/next/components/ui/sheet.tsx
slavingia: I think we should go with whatever the shadcn defaults are, and then let the designers customize from there. Most important is to finish off this migration we so we can start shipping product features without creating more tech/product debt.

rimildeyjsr:
I will resolve everything by tomorrow

apps/next/app/documents/page.tsx
raulpopadineti: I think so too. The default Shadcn grayed out description doesn't help and this is the most important content that users need to read in order to understand when their taxes are being filed. Not a blocker though. We can tweak it in a follow-up PR.

slavingia: IMO, looking at the screenshot, this feels like it should be an alert critical/destructive instead of a sheet at all

raulpopadineti: We wanted to have it always sticky to the bottom so the user sees it. I don't think the critical/destructive doesn't quite fit here since that usually means the user needs to take action on it. Whereas here, they don't have to/can't take any action on this. We're the ones filing the taxes to the IRS.

MayaRainer: Not using critical makes sense, but I think if it's an info alert at the very top people would still see it, right?

raulpopadineti: yep.

PR #137, opened 2025-04-15
refactor(ui): migrate card to shadcn card component

kr-sushil:
Migrate custom Card & related components to shadcn Card component

Sub-Issue of #18 

Before:
--

before_companysettings_card

before_roles_card

before_equity_card

After:
--

after_companysettings_card

after_roles_card

after_equity_card


apps/next/app/administrator/settings/QuickbooksIntegration.tsx
```diff
@@ -128,7 +127,7 @@ export default function QuickbooksRow() {
   };
 
   return (
-    <CardRow>
+    <div className="px-4">
```
raulpopadineti, line 130: Shouldn't be needed as it's the default in the new shadcn `Card`, no?

apps/next/app/equity/grants/DetailsModal.tsx
```diff
@@ -175,55 +176,70 @@ const DetailsModal = ({
...
-              <h3>Compliance Details</h3>
-            </CardRow>
-            {complianceDetails.map((detail, index) => (
-              <CardRow key={index} className="flex justify-between gap-4">
-                <div>{detail.label}</div>
-                <div>{detail.value}</div>
-              </CardRow>
-            ))}
+            <CardHeader>
+              <CardTitle className="font-normal">Compliance Details</CardTitle>
```
raulpopadineti, line 222: Why do we overwrite the default font? I don't think we should.

apps/next/app/equity/grants/DetailsModal.tsx
```diff
@@ -175,55 +176,70 @@ const DetailsModal = ({
...
+            </CardHeader>
+            <CardContent className="px-0">
```
raulpopadineti, line 224:
```suggestion
            <CardContent>
```

apps/next/app/equity/grants/ExerciseModal.tsx
```diff
@@ -133,75 +134,84 @@ const ExerciseModal = ({
...
+                  <div>Exercise cost</div>
+                  <div>{formatMoney(totalExerciseCost)}</div>
                 </div>
-                <div className="text-right">
-                  {formatMoney(new Decimal(optionsToExercise).mul(companySharePrice))}
-                  <br />
-                  <span className="flex justify-end text-sm">
-                    <Delta diff={equityValueDeltaPercent} />
-                  </span>
+                <HorizontalDivider />
```
raulpopadineti, line 192: Not sure if we want to keep the dividers we had between card rows as before when migrating to Shadcn. I think it would be good to use the components as close to Shadcn as possible. @slavingia @jc26 wdyt?

Gaps look also inconsistent across these cards and we probably want to fix that too.

If we want to do that, it would probably be good to make it the default and integrate that in the `CardContent` using something like:

`[&>*]:border-b [&>*]:border-black [&>*]:pb-4 [&>*]:mb-4 [&>:last-child]:border-b-0 [&>:last-child]:mb-0`

raulpopadineti (commented):
@kr-sushil _a thanks for the contribution! I've left a couple of comments.

jc26:
Discussed with @slavingia in person. @kr-sushil Scratch what I said above. This approach is fine for now. We'll probably redesign the cards down the road, but this is good for now. For the dividers, let's use [shadcn's Separator component](https://ui.shadcn.com/docs/components/separator).

raulpopadineti:
@slavingia if we get rid of the new `HorizontalDivider` component added in here along with its usages, this only requires a few minor changes. Otherwise, it's mergeable. Most of it comes from our own design inconsistencies because we used different spacings for different cards.

slavingia:
Here's a quick v0 using the separator component:

image

https://kzmnylx1nhdovqauyudy.lite.vusercontent.net

apps/next/app/administrator/settings/QuickbooksIntegration.tsx
slavingia: Ideally yeah, we should let the Card components handle these padding changes so we can change globally if/when we decide too.

kr-sushil:
@jc26 @raulpopadineti  @slavingia One question 

For all the card that look like below attached cards , should i maintain the spacing for the separator like in the design Sahil has added in the comment

card-line

kr-sushil:
No issue i can add separator side spacing in this PR only, just asked for confirmation purpose.

Will work on the comment changes.

apps/next/app/equity/grants/DetailsModal.tsx
slavingia: IMO for this one, we shouldn't use CardHeader at all. This all feels like CardContent

slavingia (approved):

PR #151, opened 2025-04-15
Fix document template seeds

MayaRainer:
Currently, the local seeds for the built-in document templates are broken. This PR fixes that.


MayaRainer:
@jc26 this fixes the issue you were having!

Issue #23, opened 2025-03-27
Add searching, sorting, filtering to tables


ershad:
[Example final result](https://v0.dev/chat/open-in-v0-mQP21zP6B1B?b=b_EavQrNLbUOQ) after all changes are in place:

<img width="1623" alt="Image" src="https://github.com/user-attachments/assets/f79f677c-ad5c-42a1-ac1b-8fc7a09b7bb3" />

(forgot total row, see [here](https://v0.dev/chat/open-in-v0-mQP21zP6B1B?b=b_picrqnJC2YF))

- [x] Add shadcn

**1. People** 

- [ ] Use https://x.com/pacovitiello/status/1879125743848095754?s=46 / https://originui.com/table for [People](https://app.flexile.com/people) table
- [ ] Remove tabbing between Onboarding, Active, Alumni
- [ ] Default to showing Active
- [ ] Filter by name...
- [ ] Allow showing and hiding columns
- [ ] Move button to Invite person (updated copy) away from header
- [ ] Click to sort

<img width="183" alt="Image" src="https://github.com/user-attachments/assets/0cb9ce0e-387b-41a0-855e-6f6b78424425" />

<img width="490" alt="Image" src="https://github.com/user-attachments/assets/14d56faa-122e-45e9-a14c-765ad8124c6d" />

<img width="212" alt="Image" src="https://github.com/user-attachments/assets/1a24db46-a01f-4f4f-82d5-c6feb5e7563d" />

<img width="191" alt="Image" src="https://github.com/user-attachments/assets/b7a38e73-5666-42ad-a83b-8f4a33495484" />

- [ ] Update styles to match (table border, heading font weight, name being slightly bolder)

**2. Invoices**

- [ ] Remove Open, History tab
- [ ] Move "Download CSV" button to above table
- [ ] Add "Filter by name..."
- [ ] Allow showing and hiding columns
- [ ] Click to sort
- [ ] Updating styling to match People; move Role to its own column
- [ ] Remove search bar from headers (moved to each table individually)
- [ ] Allow filtering by Amount Min, Max; Sent on Before, After

<img width="510" alt="Image" src="https://github.com/user-attachments/assets/de1c34f9-6582-4b9e-8574-d0c3fcb01c4c" />

Switch to badge (blue bg, green bg, no icon) for status:

<img width="68" alt="Image" src="https://github.com/user-attachments/assets/13c5e27b-c51d-4154-bb85-2b224fa44aba" />

[...]

PR #160, opened 2025-04-18
Add sorting and filtering to documents table

MayaRainer:
A part of #23.

**Before**
image
image

**After**
image
image


e2e/tests/company/document_templates.spec.ts
```diff
@@ -49,13 +54,19 @@ test.describe("Document templates", () => {
...
       }
     });
-    await page.getByRole("button", { name: "New template" }).click();
-    await expect(
-      page.getByText(
-        "By creating a custom document template, you acknowledge that Flexile shall not be liable for any claims, liabilities, or damages arising from or related to such documents. See our Terms of Service for more details.",
-      ),
-    ).toBeVisible();
-    await page.getByRole("button", { name: "Equity grant contract" }).click();
+    await page.getByRole("button", { name: "Edit templates" }).click();
```
MayaRainer, line 57: This test is currently failing because the button isn't showing up with the table being hidden for the placeholder. Wdyt we should do here?

MayaRainer (commented):
@jc26 _a could you review and push any styling changes you'd like please?

jc26:
@MayaRainer Added some minor styling changes some of which are incoming as well via https://github.com/antiwork/flexile/pull/162. Have to step out for lunch but will be back after to see there's anything else I want to change.

jc26:
@MayaRainer I'm happy with those styling changes for now. Would be nice to get more diverse data in the Documents table so that we can play with it more. I've also set the table to be default sort descending by Date. Would be nice to make sure that's of course superseded by any documents with outstanding actions.

e2e/tests/company/document_templates.spec.ts
jc26: I think we're getting rid of the table entirely no? So not even in the modal since the options in the modal are basically the table? (this screenshot isn't the pixel perfect design. just showing you to say we don't need the table)
Screenshot 2025-04-21 at 4 29 26 PM

MayaRainer: But it's possible to have multiple templates per document type, so I think we do need the table, or a similar element, and an "add template" function, right?

slavingia: IMO it makes more sense to have one template per document type. E.g. a template = a type of document. I don't see a world in needing multiple templates for a single document type.

Issue #169, opened 2025-04-22
Migrate all other inputs to use shadcn input


jc26:
Components that use legacy Input (need migration):
- [x] Input.tsx
- [ ] NumberInput.tsx
- [ ] DurationInput.tsx
- [ ] DecimalInput.tsx
- [x] ComboBox.tsx
- [ ] RangeInput.tsx
- [ ] RichText.tsx
- [ ] layouts/Main.tsx

Pages/Features that use legacy Input (need migration):
- [x] apps/next/app/settings/page.tsx (migrated)
- [x] apps/next/app/administrator/settings/Settings.tsx (migrated)
- [ ] apps/next/app/onboarding/LegalDetails.tsx
- [ ] apps/next/app/onboarding/LegalCertificationModal.tsx
- [ ] apps/next/app/onboarding/PersonalDetails.tsx
- [ ] apps/next/app/documents/page.tsx
- [ ] apps/next/app/equity/tender_offers/new/page.tsx
- [ ] apps/next/app/companies/[companyId]/administrator/onboarding/details/index.tsx
- [ ] apps/next/app/companies/[companyId]/administrator/equity_grants/new/page.tsx
- [ ] apps/next/app/administrator/settings/details/page.tsx
- [ ] apps/next/app/people/new/page.tsx
- [ ] apps/next/app/administrator/settings/StripeMicrodepositVerification.tsx
- [ ] apps/next/app/updates/team/AbsencesModal.tsx
- [ ] apps/next/app/updates/company/Edit.tsx
- [ ] apps/next/app/updates/team/Task.tsx
- [ ] apps/next/app/people/[id]/page.tsx
- [ ] apps/next/app/roles/[slug]/(public)/[id]/RolePage.tsx
- [ ] apps/next/app/roles/ManageModal.tsx
- [ ] apps/next/app/invoices/ViewList.tsx
- [ ] apps/next/app/invoices/index.tsx
- [ ] apps/next/app/company_invitations/new/page.tsx
- [ ] apps/next/app/invoices/Edit.tsx
- [ ] apps/next/app/settings/payouts/page.tsx
- [ ] apps/next/app/settings/tax/page.tsx
- [ ] apps/next/app/settings/payouts/BankAccountModal.tsx

Part of #18 

[V0 prototype for rough UX](https://v0.dev/chat/simple-form-variants-ppkzIPmCxB6)

jc26:
[...]

PR #170, opened 2025-04-22
Migrate Decimal, Number, Duration inputs to use Shadcn Input

jc26:
Part of #169


jc26:
@MayaRainer I'm thinking of renaming `DecimalInput` to `NumberInput` and have it primarily cater to integers but also have decimal support. Do you see any issue with this? Currently, `NumberInput` just feels like a simplified version of `DecimalInput`. Not sure we need both. I'd like to simplify. Wdyt?

cc: @slavingia

MayaRainer:
I'm down to merge the two - it was initially added in https://github.com/antiwork/flexile-old/pull/3777 because the handling of integers and decimals was so different, but if we can just have it be `<NumberInput decimal>` that would be nice!

jc26:
@MayaRainer using a new MacBook to push and locally git wasn't tied to jc26 or jasonchang101@gmail.com so that's why we're bumping in CLA issues with user "Jason Chang". Lmk if there's a way to retroactively change the author or fix the CLA

MayaRainer:
or we can just ignore it :D

jc26:
down for that

jc26:
@MayaRainer Let me know if this is good to merge

apps/next/app/administrator/settings/StripeMicrodepositVerification.tsx
```diff
@@ -121,28 +122,39 @@ const StripeMicrodepositVerification = () => {
...
-              min="0"
-              placeholder="0"
-              label="Amount 2"
-              prefix="$"
-              help={errors.get("secondAmount")}
-            />
-          </>
+          <div className="grid gap-4">
+            <div>
+              <Label htmlFor="amount-1">Amount 1</Label>
```
MayaRainer, line 127: It would be good to migrate these to shadcn forms when we touch them instead of adding custom labels/IDs/error rendering. Can merge as is if you'd like since it's already written, but worth fixing I think to help AI write better code - if you don't want to I can do it after I'm done with my current tasks, but if you do you can look at #165 for inspiration

apps/next/app/administrator/settings/equity/page.tsx
```diff
@@ -162,30 +166,75 @@ export default function Equity() {
...
+                value={conversionSharePriceUsd ?? null}
+                onChange={(value) => setConversionSharePriceUsd(value ?? 0)}
+                invalid={errors.has("conversionSharePriceUsd")}
+                prefix="$"
+                minimumFractionDigits={2}
+                decimal
+              />
+            </div>
+            <div className="grid gap-2">
+              <Label htmlFor="minimum-allocation">Minimum allocation</Label>
```
MayaRainer, line 206: How come we have more fields here now?

apps/next/app/companies/[companyId]/administrator/onboarding/details/index.tsx
```diff
@@ -22,7 +22,7 @@ const formSchema = z.object({
     street_address: z.string().min(1, "This field is required"),
     city: z.string().min(1, "This field is required"),
     state: z.string().min(1, "This field is required"),
-    zip_code: z.string().regex(/^\d+$/u, { message: "Enter a valid ZIP code" }),
+    zip_code: z.string().regex(/^\d{5}(?:-\d{4})?$/u, { message: "Enter a valid ZIP code" }),
```
Comment (MayaRainer) [inline apps/next/app/companies/[companyId]/administrator/onboarding/details/index.tsx:25]: Are you sure about this change? IIRC we previously removed this validation as non-US ZIP codes don't follow this format.

apps/next/components/NumberInput.tsx
```diff
@@ -1,30 +1,139 @@
...
+        onChange={handleChange}
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+        inputMode={decimal ? "decimal" : "numeric"}
+        className={inputClasses}
+        aria-invalid={invalid}
+        {...props}
+      />
+
+      {prefix ? (
```
MayaRainer, line 125: This feels like it should live in the `Input` in general rather than just `NumberInput`. Wdyt?

apps/next/components/NumberInput.tsx
```diff
@@ -1,30 +1,139 @@
...
-  const [input, setInput] = useState(value?.toString() ?? "");
-  useEffect(() => setInput(value?.toString() ?? ""), [value]);
+  onBlur?: React.FocusEventHandler<HTMLInputElement>;
+  onFocus?: React.FocusEventHandler<HTMLInputElement>;
+  prefix?: string;
+  suffix?: string;
+  invalid?: boolean;
+  decimal?: boolean;
+  maximumFractionDigits?: number;
+  minimumFractionDigits?: number;
```
MayaRainer, line 33:
```suggestion
  decimal?: boolean | { maximumFractionDigits?: number; minimumFractionDigits?: number; }
```

apps/next/components/NumberInput.tsx
```diff
@@ -1,30 +1,139 @@
...
+  id,
   ...props
 }: {
   value: number | null;
   onChange: (value: number | null) => void;
-} & Omit<React.ComponentProps<typeof Input>, "value" | "onChange" | "inputMode">) => {
-  const [input, setInput] = useState(value?.toString() ?? "");
-  useEffect(() => setInput(value?.toString() ?? ""), [value]);
+  onBlur?: React.FocusEventHandler<HTMLInputElement>;
+  onFocus?: React.FocusEventHandler<HTMLInputElement>;
```
MayaRainer, line 27: These should not be needed as they're already a part of the `ComponentProps`

apps/next/components/NumberInput.tsx
```diff
@@ -1,30 +1,139 @@
...
+  prefix?: string;
+  suffix?: string;
+  invalid?: boolean;
```
MayaRainer, line 30: Should be handled by `Form` rather than manually I think (though we can keep it around in the short term if that would be too much work to fix rn)

e2e/tests/company/administrator/edit-contractor.spec.ts
```diff
@@ -149,7 +149,11 @@ test.describe("Edit contractor", () => {
     await page.getByRole("button", { name: "Create new" }).click();
 
     await page.getByLabel("Name", { exact: true }).fill("Example Role");
-    await page.getByLabel("New role").getByLabel("Rate", { exact: true }).fill("200");
+
+    const rateInput = page.getByRole("dialog", { name: "New role" }).locator("#pay-rate");
```
MayaRainer, line 153: Could we use `withinModal` here?

MayaRainer (commented):
@jc26 _a added some comments - happy to address any you don't/can't get to yourself after I'm done with my current tasks, just lmk!

slavingia:
This should be merged. I think we can move a lot faster by screen sharing, resolving inline, and then merging. The async PR loop is very slow atm.

apps/next/components/NumberInput.tsx
jc26: Are you talking about just the onFocus and invalid? Or including prefix and suffix. Re: your previous comment, I agree, I think the prefix and suffix should live in input.tsx

apps/next/app/administrator/settings/equity/page.tsx
jc26: Not sure .. Removing.

jc26:
Merging and addressing @MayaRainer's comments in a followup

slavingia:
She can address them. The point now is to ship as much as possible. Once the product debt is resolved, we can worry about the tech debt.

IMO: full-time, in-office ship maximum features collaboratively. Use 3rd party contractors to do lower priority things such as reduce tech debt

apps/next/components/NumberInput.tsx
jc26: Agreed

PR #195, opened 2025-04-25
Don't run lints on Vercel deploys

MayaRainer:
Currently, linting errors block Vercel deployments. This PR fixes that.


slavingia:
If we're okay with a lint being deployed, why are we enforcing it locally? Shouldn't it be the opposite, we should be more okay with local development but not okay with lints breaking if they're being deployed.

MayaRainer:
The lints are already covered by Autofix so it's not technically needed on Vercel as well. That said, I generally agree that code that breaks lints should not be merged, i.e. ideally we always wait for autofix and status checks to run and pass before merging. Happy to revert if you prefer.

PR #229, opened 2025-04-28
Move Equity tab in settings/equity to Invoices

jc26:
## Why

- Doesn't make sense for Equity to be separate consideration from Invoices
- Cut down on the number of tabs that we have. We're moving to phase out tabs app-wide.

## Before
Screenshot 2025-04-28 at 4 20 35 PM
Screenshot 2025-04-28 at 4 20 56 PM

## After
Screenshot 2025-04-28 at 4 19 40 PM


jc26:
Taking a crack at updating the tests and then I'll assign to @MayaRainer

jc26:
@MayaRainer Actually, I'll let you handle it. The AI sweeping changes to the tests might hurt more than help.

jc26:
FYI This is what project based looks like

Screenshot 2025-04-28 at 4 45 59 PM

jc26:
Empty state for the Invoices table has been removed since the quick invoice section basically functions as an empty state. FYI @slavingia approved designs IRL

slavingia (approved):
Looks awesome!

MayaRainer:
Will get to this tomorrow - fixing tests in #208 took a bunch of time.

jc26:
@MayaRainer Since we don't have the `Save equity split` button anymore, could we make it so that the first invoice sent with a new equity split effectively saves the equity split?

slavingia:
> @MayaRainer Since we don't have the `Save equity split` button anymore, could we make it so that the first invoice sent with a new equity split effectively saves the equity split?

Def think we should do that!

MayaRainer:
One thing that's confusing to me about this is that there doesn't seem to be a way to update your equity split on the full new invoice page. (The code also seems broken in general in that it doesn't save the selected equity split at all, but I can fix that) Are we thinking we might merge the two ways of submitting invoices, or should we display that selector there as well?

I also think it might be a bit confusing to always display this selector even when the allocation is locked and it can't be changed (which would be most of the time). Maybe we should replace it with the alert, or at least more obviously disable it?

Happy to chat about this on a call later/tomorrow if that's easier

jc26:
@MayaRainer Agree that it should be always present but enabled/disabled! I was thinking of adding some helper text to the range input so that there's some context when it is disabled, like you said most of the time. And when it's not disabled, we should also have helper text explaining that it's time to select a new equity split which will be locked for the rest of the year. 

Yea let's add the range input selector into the other new invoice page as well. We can probably open another PR where we look into just killing that page entirely as I don't think it's necessary honestly. But one big change at a time 🙂 Wdyt?

slavingia:
I think the quick invoice page could transform into the new invoice page upon focus. The table would hide, new invoice button would hide, and Invoices could fade into New invoice. Would be neato. Can be done separately

jc26:
@slavingia noted. added to my backlog

raulpopadineti:
Will fix all tests and merge.

slavingia:
resolve merge conflicts

slavingia:
nvm

PR #238, opened 2025-04-30
Fix DocuSeal embed dialog height in tests

raulpopadineti:
### Before
image

### After
image


raulpopadineti:
@jc26 looks like this makes CI happy.

PR #242, opened 2025-05-01
Migrate to accessible date picker

jc26:
## Before
Screenshot 2025-05-01 at 12 32 57 PM

## After
Screenshot 2025-05-01 at 12 29 40 PM


jc26:
@ershad is trying to help me fix some UTC. Issues. The backend saves the right date, but the frontend date displays the current date - 1. @raulpopadineti you bumped into this yesterday right? Can't remember your fix for it. This time, it seems like it's failing on GH servers as well.

raulpopadineti:
@jc26 yeah, the problem was that we were doing things a bit differently before because we only took the date part of the timestamp and format it on the frontend, whereas now, we're including the timestamp which messes up the formatting when in a different timezone than UTC which is the timezone we store timestamps in our DB.

apps/next/components/ui/input.tsx
```diff
@@ -1,20 +1,22 @@
...
 import { cn } from "@/utils/index";
 
-type InputProps = React.ComponentProps<"input">;
-function Input({
-  className,
-  type,
-  value,
-  ...props
-}: Omit<InputProps, "value"> & { value?: InputProps["value"] | null }) {
+type InputProps = Omit<React.ComponentProps<"input">, "value" | "onChange"> & {
```
MayaRainer, line 4: I don't really understand the changes in this file - they seem to all be no-ops to me.

Lmk if they were an AI hallucination and we can revert them.

e2e/helpers/index.ts
```diff
@@ -4,3 +4,6 @@ export const selectComboboxOption = async (page: Page, name: string, option: str
   await page.getByRole("combobox", { name }).click();
   await page.getByRole("option", { name: option, exact: true }).first().click();
 };
+
+export const fillDatePicker = async (page: Page, name: string, value: string) =>
```
MayaRainer, line 8: A bit awkward, but didn't find a better way to fill the date picker input. There's [a related discussion here](https://github.com/adobe/react-spectrum/discussions/7136) without a proper solution either.

apps/next/app/equity/tender_offers/new/page.tsx
```diff
@@ -51,10 +53,12 @@ export default function NewBuyback() {
         },
       });
 
+      const localTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+
       await createTenderOffer.mutateAsync({
         companyId: company.id,
-        startsAt: new Date(startDate),
-        endsAt: new Date(endDate),
+        startsAt: startDate.toDate(localTimeZone),
```
MayaRainer, line 60: since we seem to store this as a date time, maybe we should also display that time to users in the UI and email? otherwise, might make the most sense to just store the date (as UTC) and not have to deal with timezones at all

MayaRainer (commented):
@jc26 fixed the tests and added a few comments! should be good to merge once the changes in input are (probably) reverted

e2e/helpers/index.ts
jc26: yea I would've never reached that on my own. thank you!

apps/next/app/equity/tender_offers/new/page.tsx
jc26: Agreed that it makes the most sense just to store it as date. If you're picking a future date which is most likely the case, wouldn't make sense that we grab your local time + the future date you picked. Unless we add a time picker as well, which is unnecessary

apps/next/components/ui/input.tsx
jc26: Reverted some changes and moved all `file:` styles into `type === "file" &&`. Let me know if you have a better solution

PR #251, opened 2025-05-02
Feature/migrate radiobutton to shadcn

itsxdamdam:
Update RadioGroup to use RadioGroupPrimitive


MayaRainer:
I'll close this as it doesn't really make sense to use the radix radio buttons with how we're using them. Maybe we can get rid of `RadioButtons` entirely in a follow-up.

itsxdamdam:
Ok. Also could you provide more clarity on installing and using shadcn components within the project. I'm working on a pr for the select component

MayaRainer:
As mentioned in #18, it's "Easiest to get rid of Input, FormSection, Select, Label, and the old way of doing forms/validation all at once than trying to migrate fifty usages at once, and avoids conflicts between them." At this point it probably doesn't make sense to migrate form components separately, but rather form-by-form. It may be easiest for you to just focus on another task for now.

PR #264, opened 2025-05-05
Migrate invoice date pickers

jc26:
Invoice quality of life updates

## What

- migrate date pickers in invoice creation

## Screengrabs
https://github.com/user-attachments/assets/ee2ba5e0-df20-4fd8-b091-51f8bcb04edf

Screenshot 2025-05-05 at 2 08 31 PM


apps/next/app/invoices/Edit.tsx
```diff
@@ -207,6 +225,9 @@ const Edit = () => {
       });
       await trpcUtils.invoices.list.invalidate({ companyId: company.id });
       await trpcUtils.documents.list.invalidate();
+      if (!id) {
+        toast.success("Invoice created successfully");
```
slavingia, line 229: Don't think we should rely on toasts, but the invoice showing up in the table view. Similar to native Mac software, there generally aren't tooltips and alerts/toasts.

slavingia (changes requested):

apps/next/app/invoices/Edit.tsx
jc26: Ahh got it. We'll just do the date pickers in this PR then

MayaRainer:
Will fix tests and get this merged today.

PR #265, opened 2025-05-05
Update validation error message for contractor invite in `invite_worker_spec.rb`

raulpopadineti:
Fixes the [latest failing CI error](https://github.com/antiwork/flexile/actions/runs/14843773422).

Regression from #259.


raulpopadineti:
cc @MayaRainer

PR #277, opened 2025-05-07
Replace `formatDate` with `formatServerDate` for dates in BuybackView

raulpopadineti:
Fix showing incorrect dates for buy backs where timezone pushes the timestamp back in time.

### Before
image

### After
image


jc26:
@raulpopadineti Cool! I can convert it to a nicer looking date in a followup PR?

PR #305, opened 2025-05-16
Full width layout

slavingia:
Updates the main layout to be full width

## Summary by CodeRabbit

- **Style**
  - Removed maximum width constraints from the header and main content containers for improved layout flexibility.


laugardie (approved):
Love it! I was thinking of doing the same and dropping the table margins, but I’ll hold off on the margins for now. Gonna shape up the settings a bit first so they’re not totally broken.
@slavingia _a

PR #469, opened 2025-07-09
Convert auth pages to use layout.tsx pattern

CyMule:


frontend/components/layouts/Settings.tsx
```diff
@@ -1,3 +1,5 @@
+"use client";
```
MayaRainer, line 1: Could we just inline this file into `settings/layout`?

frontend/app/(auth)/layout.tsx
```diff
@@ -17,5 +17,11 @@ export default function AuthLayout({ children }: { children: React.ReactNode })
     }
   }, [user, searchParams]);
 
-  return children;
+  return (
+    <div className="flex h-full flex-col">
```
MayaRainer, line 21: Why not keep using `SimpleLayout` here?

MayaRainer (commented):
@CyMule _a added two comments!

slavingia:
Relatively small changes await so would love to merge this in tomorrow.

slavingia:
There's a good chance that doesn't get merged due to lack of code quality (haven't investigated closely) so always keep that in mind. First who finishes wins, not first who starts.

PR #473, opened 2025-07-09
Tweaks to monthly dividend report CSV

sharang-d:
Changes to #431

Sample CSV:
Screenshot 2025-07-09 at 1 25 10 PM


backend/app/services/dividend_report_csv.rb
```diff
@@ -23,48 +20,31 @@ def generate
...
+          dividend_round.company.name,
           total_dividends,
           flexile_fees,
           total_transfer_fees,
           total_ach_pull,
-          successful_dividends.count,
-          round.status,
-          payment&.processor_name,
-          payment&.transfer_id
+          dividends.count,
```
sharang-d, line 45: The column name for this cell is "Number of investors" so this should be `dividends.select(:company_investor_id).distinct.count` but I'm not sure if we need the investors count so I let it be (we can change the column header to say 'Number of dividends' instead).

sharang-d (commented):
@slavingia _a ready for review/comments. (Didn't wait for tests to run since they may need approval to run while I don't have write access)

sharang-d:
There's also this class for dividend reporting per payment (vs per ACH pull): https://github.com/antiwork/flexile/blob/main/backend/app/services/dividend_payment_csv.rb

slavingia:
I think some stuff got into this branch that shouldn't be. could you update with main?

sharang-d:
I updated the branch with main as autofix failed. Now this branch is ahead of the base branch i.e. https://github.com/antiwork/flexile/pull/431 which is why extra code is showing. If you update that PR's branch with main via the button on the UI, the diff here should look good. If not, I will recreate the commits here.

slavingia:
got it. seeing some conflicts now?

sharang-d:
Fixed conflicts, and the diff looks good again.

Issue #494, opened 2025-07-10
New contractor bugs


slavingia:
- Search for role doesn't work
- Tapping outside doesn't hide
- Selecting should auto-hide
- Past values aren't being defaulted to (role, pay rate)

chaitanyya:
Okay these issues i can reproduce, unrelated to the change i made to fix the issue.

chaitanyya:
I think @jeremyosih fixed one of the issue, rest of the UX issues are included in my PR, great co-incidence

chaitanyya:
Awesome so both PRs address the issues listed here.

PR #513, opened 2025-07-11
Fixes bug where new companies cannot enter contractor role.

the-wunmi:
The UX for adding role to a contractor was optimized on #494 , but introduces a bug where a company cannot type in roles not part of the existing roles list, causing new companies without exising workers with roles to not be able to add roles to contractors they create, this PR addresses this, alongside fixes to the related failing tests in CI.

# Before

https://github.com/user-attachments/assets/81157156-2ddc-4801-b8b7-db94943cae61

# After

https://github.com/user-attachments/assets/ea956494-34c5-4765-bd60-7446d9fa57c6

# e2e tests

image

## Summary by CodeRabbit

* **New Features**
  * The ComboBox now supports an optional input event handler, allowing dynamic updates based on user input.
  * Users can add new roles directly by typing them into the ComboBox, which will then appear as selectable options.

* **Bug Fixes**
  * Improved test reliability by updating how role selection is performed in end-to-end tests, ensuring accurate simulation of user interactions with the ComboBox.

* **Refactor**
  * Updated tests to interact with the role field as a ComboBox rather than a simple text input, aligning tests with the current UI behavior.


slavingia:
Maybe this works actually, though the empty state is too short. 

Can we make it "Search or create role..."

slavingia:
Could you confirm all (relevant) tests pass in your branch?

slavingia:
Because they require secrets

the-wunmi:
Updated the search placeholder to "Search or create role...", and validated that related tests passes. @slavingia 

# label
image

# e2e
image

slavingia:
@laugardie any thoughts on how we could make this a bit less janky?

slavingia:
Reverting to original + bug fixes instead cc @MayaRainer

PR #535, opened 2025-07-13
Refactor test auth flow to dynamically create Clerk users

divanshu-go:
- Removed hardcoded Clerk test users
- Dynamically creates Clerk users via API during test setup
- Updated `setClerkUser` and `login` helpers to support this flow
- Simplified email generation in user factory
- Ensured sign-in and sign-out flows work with the updated setup

image
image

## Summary by CodeRabbit

* **Refactor**
  * Streamlined test user management by removing hardcoded test users and simplifying setup and cleanup logic.
  * Updated helper functions to work with full user objects instead of just user IDs.
  * Improved user creation for tests by adjusting email formatting and adding sign-in timestamp fields.

* **Tests**
  * Simplified login tests by using direct sign-in methods, reducing manual form interactions.

### DIFF
--- a/e2e/factories/users.ts
+++ b/e2e/factories/users.ts
@@ -17,7 +17,7 @@ export const usersFactory = {
     const [user] = await db
       .insert(users)
       .values({
-        email: faker.internet.email().toLowerCase(),
+        email: `${faker.internet.email().split("@")[0]}+clerk_test@example.com`.toLowerCase(),
         legalName: faker.person.fullName(),
         preferredName: faker.person.firstName(),
         encryptedPassword: await bcrypt.hash("password", 10),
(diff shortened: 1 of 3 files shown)


divanshu-go:
@slavingia This refactor simplifies the test setup by removing the need for manually maintaining the four hardcoded Clerk test users.

Also, regarding [#435](https://github.com/antiwork/flexile/issues/435), would you like me to start working on that?

MayaRainer:
I don't think we should be making this change. We initially hardcoded the Clerk users because the users we create here would persist beyond the scope of the test, and Clerk has a (relatively low) limit of users for the dev environment. If something changed since then lmk!

PR #577, opened 2025-07-17
Optimized logged_in_user (antiwork#560)

OxJustCode:
eager loads, preload roles & add tax info to legal details


slavingia:
Please include specific performance benchmarks for any optimizations.

PR #539, opened 2025-07-14
Fix wrong options value percentage in `ExerciseModal`

wzulfikar:
The % indicator for "Options value" seems unexpectedly high and I suspect it's not intentional (i.e. a bug). 

We passed cent-based number into `Delta` component which formats the number using toLocaleString with `style: "percent"`. This causes the number to be multiplied by 100 again, causing it to render `16,465.12%` instead of `164.65%`.

How to replicate:
1. Log in to Flexile
2. Go to https://flexile.com/equity/options (make sure you have vested options available for exercise)
3. Click "Exercise Options" to show the modal

image

image

<hr/>

**Preview:**

| Before | After |
|--------|-------|
| image    |     image  |

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the equity value delta display from a percentage to a decimal fraction in the exercise modal.
  * Adjusted the modal to show the correct equity value delta format and updated related labels.

* **Tests**
  * Enhanced end-to-end tests with updated expected values and added explanatory comments for clarity.

PR #604, opened 2025-07-21
Remove min-h-screen from `ExerciseModal`

wzulfikar:
Was checking #539 and I noticed the ExerciseModal has changed from sidebar to dialog. While it was a sidebar, it made sense to have `min-h-screen` but with the current style, the `min-h-screen` creates extra vertical space.

Removing the `min-h-screen` should fix it. 
- The dialog still have vertical scroll when the window's height is less than the modal
- The UI still look proper in mobile screen too.

https://github.com/user-attachments/assets/9be06250-a62c-46ab-8f8f-6466cd555a4e

## Summary by CodeRabbit

* **Style**
  * Updated modal styling to remove enforced minimum screen height, allowing for more flexible modal sizing.


slavingia:
It should be a sidebar still. I think the best fix is to fix that. On mobile it can be a sheet that comes up from the bottom.

wzulfikar:
I see, so prob a regression. Might look deeper into it in weekend.

slavingia:
Yeah think so

slavingia:
Keeping for now :)

wzulfikar:
Sure 🫡

Issue #569, opened 2025-07-16
Option to download invoice PDF


Udit-takkar:
Background:
You have the option to create invoice in flexile but there should be an option to download them as well as i have to need upload the invoice on other platforms like https://www.skydo.com/ for withdrawal.

Current Behavior:
No download button for invoice created on flexile.com

Expected Behavior:
I want the ability to download invoice pdf

Additional Context:
Other websites like deel.com give the option to download invoice pdf

Udit-takkar:
I can work on this issue if i can get a figma design of the invoice pdf

slavingia:
I think you should just be able to do Apple P and it should run the proper print styling. We could add a print button, which would effectively do the same thing.

Udit-takkar:
@slavingia well if you do CMD + P then the invoice does not look very professional and genuine and if you have a big description then it just doesn't display the complete description.

![Image](https://github.com/user-attachments/assets/a7e19e6b-3bd1-4517-a164-bc94925c568b)

slavingia:
> @slavingia well if you do CMD + P then the invoice does not look very professional and genuine and if you have a big description then it just doesn't display the complete description.
> 
> ![Image](https://github.com/user-attachments/assets/a7e19e6b-3bd1-4517-a164-bc94925c568b)
> 
> 

Worth fixing this then I think as part of this task.

log-xp:
Hey @slavingia  , 

I worked some time on this , Apple P  and css only . Will you be intrested or this is not planned .
  - page.tsx: Added PrinterIcon import, Print button, and invoice-print class
  - globals.css: Added comprehensive print styles with clean, professional formatting


UI :

<img width="2366" height="892" alt="Image" src="https://github.com/user-attachments/assets/c565d7c4-2366-4407-84ed-67f63a4a797c" />

PDF :
<img width="463" height="652" alt="Image" src="https://github.com/user-attachments/assets/5a9248f3-d2a5-49f8-870a-639e10bcdcb6" />

log-xp:
Hey @slavingia , 

sorry to add again , can we do this ? Its a css solution , minimal code . If you are interested I will raise a PR .

PR #610, opened 2025-07-22
feat: add download invoice functionality

Udit-takkar:
Fixes https://github.com/antiwork/flexile/issues/569

TODO
- [ ] Review and improve code

Before:-
When you use `CMD + P` to download invoice from the page 
Screenshot 2025-07-23 at 2 20 09 AM

After:-
You click on download button and you get a pdf file like below

Screenshot 2025-07-23 at 1 51 25 AM


Udit-takkar:
@slavingia i am almost complete. i just want to confirm are you okay with using this approach?

slavingia:
Should be CSS only, not so many new lines of code / new components that need to be checked with every change

PR #646, opened 2025-07-27
Remove remaining parts of onboarding

MayaRainer:
Closes #454.


slavingia:
Awesome! We'll just want to ensure the onboarding checklist shows up nicely for investors (especially) and workers so that people get paid their dividend. The existing flow is nice as it forces all of that info in a linear way which works for that single use case. This is slightly worse but the checklist in the nav and good alerts/notices should provide the same experience with a lot less custom UI.

MayaRainer (commented):
Remaining test failures look unrelated, I can address those separately if they're still happening after my reviews

slavingia:
LGTM. Small thing may be to re-order these from completed to not completed, but we can do that later along with some cleanup.

image

PR #675, opened 2025-07-30
feat: svgr

nazargladish:
* This PR adds SVGR to render SVGs as JSX. I needed this while working on the #671, using `<Image />` for small icons caused noticeable lag under network and CPU throttling. Inlining them as JSX made the UI much more responsive.

* We can still import SVGs as static URLs using `?url`, so this change just extends functionality without removing anything.

**UX impact (slow network):**

|  |  |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/c05b2aab-7e89-499a-8f85-a736f979737b">  | <video src="https://github.com/user-attachments/assets/76da73c2-f769-4531-acc9-e097b1a76552">|

## Summary by CodeRabbit

* **New Features**
  * Improved SVG handling allows SVGs to be imported either as React components or as image URLs, providing greater flexibility in asset usage.

* **Refactor**
  * Updated logo and icon imports throughout the app to use SVGs as React components where appropriate, simplifying icon rendering and reducing reliance on the Next.js Image component.

* **Chores**
  * Added TypeScript declarations for SVG imports.
  * Updated TypeScript configuration to include new type declarations.
  * Added the "@svgr/webpack" development dependency for enhanced SVG support.

### DIFF
--- a/frontend/app/(dashboard)/layout.tsx
+++ b/frontend/app/(dashboard)/layout.tsx
@@ -48,7 +48,7 @@ import {
   useSidebar,
 } from "@/components/ui/sidebar";
 import { useCurrentCompany, useCurrentUser, useUserStore } from "@/global";
-import defaultCompanyLogo from "@/images/default-company-logo.svg";
+import defaultCompanyLogo from "@/images/default-company-logo.svg?url";
 import { storageKeys } from "@/models/constants";
 import { trpc } from "@/trpc/client";
 import { request } from "@/utils/request";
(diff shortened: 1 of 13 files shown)


slavingia:
Showing before/after of the UX would be great

MayaRainer:
You said that this is specific to `Image`, but icons don't use that, and intuitively the icon library we're using already would use React components out of the box - could you lmk what I'm missing here please?

Issue #506, opened 2025-07-11
Fix Dev Setup Conflicts and Improve Onboarding Documentation


wiksien:
## What

While doing a fresh development setup on Linux Mint, I noticed several issues that could be improved in `README.md`, `bin/setup`, and `bin/dev` to make local development setup easier for new contributors.

<details>
<summary>More in depth description</summary>

### Current Issues

1. **Homebrew PostgreSQL conflicts with Docker PostgreSQL**
   If the user runs `bin/setup`, PostgreSQL is installed via Homebrew and started locally on port 5432. This then conflicts when `bin/dev` is started, as `bin/dev` uses a Docker instance of PostgreSQL that is a separate instance but tries to access the same port 5432 and errors out. I think the Homebrew instance is by mistake - I don't see how it makes sense given the app uses the Docker instance. Additionally, given the current setup, the `CREATE USER username WITH LOGIN CREATEDB SUPERUSER PASSWORD 'password';` command only affects the Homebrew instance of PostgreSQL, not the Docker instance. This might be the cause of the common issue described in the `README.md`.

2. **Ruby installation process is incomplete**
    For rbenv to install correctly on a fresh machine, it needs additional system dependencies. This is not handled in `bin/setup`.

3. **Missing SSL certificate setup**
    `mkcert` is required for local HTTPS development but not mentioned in setup.

4. **Unclear environment configuration**
    The current documentation is a little bit misleading for developers unaware of the fact that the Vercel .env pull is only for people who already have access to the company Vercel project. Otherwise, all they can do is update .env on Vercel in their own project. The [Shortest repository](https://github.com/antiwork/shortest) provides a good example of how this could be better documented.

5. **Lack of a guide for setting up .env keys**
    At least for the basic keys required for setup, it would be good to have documentation in `README.md` on their setup. The [Shortest repository](https://github.com/antiwork/shortest) also provides a good example of how this could be done.

6. **Missing bundler install**
    `bin/setup` is missing a `gem install bundler` before installing foreman.

7. **Database readiness race condition in bin/dev**
    The current `bin/dev` script runs `bin/rails db:prepare` immediately after starting Docker services, but before PostgreSQL is ready to accept connections. This causes database connection failures. The script should wait for the database to be ready before running database operations.

8. **Redis connection timing issues**
    The `README.md` mentions Redis connection failures on port 6389, suggesting similar timing issues with Redis startup that could be resolved with proper service coordination.

</details>

## Why

[...]

PR #725, opened 2025-08-02
Add warning about using windows and wsl for setup in README.md

wiksien:
Related to #506 

image

## Why

Windows and WSL had been prone to problems when it comes to DX experience, like complicated mkcert configuration in WSL and other issues. Since we don't plan to support Windows based DX, it is best that we explicitly recommend users to use either Linux or MacOS in order to avoid frustration by new contributors when setting up their local, that can lead to giving up on contributing.

## Additional Notes

The PR aims to not exclude Windows based contributors by giving the option of making a VM instance of Linux.

## Summary by CodeRabbit

* **Documentation**
  * Added a warning in the Setup section recommending MacOS or Linux for setup and advising Windows users to use virtualization tools due to known issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Contractor payments as easy as 1-2-3.
 ## Setup
 
 > **⚠️ Warning:**  
-> We highly recommend using either a MacOS system or a Linux distribution for setting up Flexile on your machine. Using Windows and WSL is known to cause a lot of issues that are hard to debug. If you are using Windows, consider setting up an instance of Linux distro via tools like [VirtualBox](https://Virtualbox.org).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,9 @@ Contractor payments as easy as 1-2-3.
 
 ## Setup
 
+> **⚠️ Warning:**  
+> We highly recommend using either a MacOS system or a Linux distribution for setting up Flexile on your machine. Using Windows and WSL is known to cause a lot of issues that are hard to debugg. If you are using Windows, consider setting up an instance of Linux distro via tools like [VirtualBox](https://Virtualbox.org).
```
Anshgrover23, line 8:
```suggestion
> We highly recommend using either a MacOS system or a Linux distribution for setting up Flexile on your machine. Using Windows and WSL is known to cause a lot of issues that are hard to debug. If you are using Windows, consider setting up an instance of Linux distro via tools like [VirtualBox](https://Virtualbox.org).
```

slavingia:
This is true of all OSS so don't think it needs to take up important README space.

PR #739, opened 2025-08-04
fix broken invite flow

kanhayaKy:
The removal of the pre-onboarding state has broken the invite link flow. Previously, the AcceptInviteService was triggered while the user was in the onboarding state, which ensured that a CompanyWorker was created. Now that the onboarding state has been removed, the CompanyWorker is no longer being created when the invite is accepted.

REF : https://github.com/antiwork/flexile/pull/646/files#diff-48e07e012c5e026a033affd66353fc033b09533159ce6a3de733235b681024dc


slavingia:
Any e2e specs that should be added/edited to catch this next time?

kanhayaKy:
@slavingia Updated the relevant test
image

PR #801, opened 2025-08-08
Update Docker configuration to use nginx instead of OpenResty

ershad:
Replaced the OpenResty image with the nginx:alpine image in both local Docker Compose files. Removed the command for installing LuaRocks and running OpenResty, simplifying the setup for local development.

Thanks to @sharang-d for the ping about the issue with luarocks.

## Summary by CodeRabbit

* **Chores**
  * Updated the Docker Compose configuration to use the standard nginx image instead of OpenResty for local development environments.
  * Simplified the nginx service startup by removing custom commands and package installations.

### DIFF
--- a/docker/docker-compose-local-linux.yml
+++ b/docker/docker-compose-local-linux.yml
@@ -21,16 +21,13 @@ services:
       - ./tmp/redis:/data
 
   nginx:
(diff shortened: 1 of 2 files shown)


sharang-d:
Thanks, I'll test it out soon.

Issue #792, opened 2025-08-07
User impersonation by admins


MayaRainer:
Now that Clerk is gone, we no longer have a way of impersonating users on Flexile to debug issues. Would be good to re-add as it saves a lot of time doing support.

Add visually via admin UI.

slavingia:
Cancelled clerk to save $225 a month :D

namanxajmera:
@MayaRainer I've implemented the user impersonation feature you requested in PR #830.

## What's Delivered

**Simple Console Command** (as requested by @slavingia):
```bash
rails impersonation:generate_url[user@example.com]
# Output: https://flexile.co/admin/impersonate?token=AbCdEf...
```

**Key Features:**
- ✅ **Admin-only access** - Uses existing `team_member?` constraint
- ✅ **Secure** - 5-minute expiring signed tokens 
- ✅ **Simple workflow** - Console command → paste URL → impersonate user
- ✅ **Stop anytime** - DELETE endpoint to restore original admin user
- ✅ **JWT compatible** - Extends existing auth to support browser cookies

## Why This Approach

This delivers the **MVP console command approach** @slavingia suggested instead of the complex web UI in PR #797. It's:
- **Minimal** - <100 lines of core logic vs 200+ 
- **Secure** - Uses Rails signed IDs + existing admin routes
- **Fast to implement** - Extends current JWT auth instead of new systems

## Usage

1. Admin runs: `rails impersonation:generate_url[user@example.com]`
2. Pastes generated URL in browser
3. System switches admin to impersonate that user  
4. Admin can debug issues as the user
5. Visit `/admin/impersonate` (DELETE) to stop impersonation

The implementation includes comprehensive tests, documentation, and error handling for all edge cases.

Ready for review\! This should solve your support debugging needs with minimal complexity.

AJ0070:
> @MayaRainer I've implemented the user impersonation feature you requested in PR #830.
> 
> ## What's Delivered
> 
> **Simple Console Command** (as requested by @slavingia):
> ```bash
> rails impersonation:generate_url[user@example.com]
> # Output: https://flexile.co/admin/impersonate?token=AbCdEf...
> ```
> 
> **Key Features:**
> - ✅ **Admin-only access** - Uses existing `team_member?` constraint
> - ✅ **Secure** - 5-minute expiring signed tokens 
> - ✅ **Simple workflow** - Console command → paste URL → impersonate user
[...]

PR #805, opened 2025-08-08
feat: impersonate user

Udit-takkar:
Fixes: https://github.com/antiwork/flexile/issues/792

https://github.com/user-attachments/assets/c6365816-b58b-4a63-978e-00fbd0a243cd


Udit-takkar:
UI not needed

PR #817, opened 2025-08-10
Flexile fee calculator

nazargladish:
## Why

- Centralize fee calculation on the backend since it’s required in multiple places.
- Move [this function](https://github.com/antiwork/flexile/pull/767/files#diff-6e27f621f86322f271385b49b01c93663f4016dcd05269ea9cc761a03c04c448R235-R242) to the backend, where the business logic belongs.

## Testing Notes
Screenshot 2025-08-10 at 12 44 12
Screenshot 2025-08-10 at 12 45 14
Screenshot 2025-08-12 at 18 23 57

## Summary by CodeRabbit

* **New Features**
  * Centralized fee calculator service for invoice and dividend fee computation, now used platform-wide.

* **Bug Fixes**
  * Consistent fee logic with correct rounding, base fees, percentage application and maximum caps; validations updated to use the centralized values.

* **Tests**
  * Added unit tests covering zero, small, large (capped), and rounding scenarios for fee calculations.


backend/app/services/flexile_fee_calculator.rb
```diff
@@ -0,0 +1,31 @@
...
+    percentage_fee_cents = (total_amount_in_usd_cents.to_d * INVOICE_PERCENTAGE.to_d / 100).round
+    fee_cents = INVOICE_BASE_FEE_CENTS + percentage_fee_cents
+    [fee_cents, INVOICE_MAX_FEE_CENTS].min
+  end
+
+  def calculate_dividend_fee_cents(total_amount_in_cents)
+    percentage_fee_cents = (total_amount_in_cents.to_d * DIVIDEND_PERCENTAGE.to_d / 100).round
+    fee_cents = DIVIDEND_BASE_FEE_CENTS + percentage_fee_cents
+    [fee_cents, DIVIDEND_MAX_FEE_CENTS].min
+  end
```
sharang-d, line 30: These can be class methods instead of doing new + method.

backend/spec/services/flexile_fee_calculator_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
```
sharang-d, line 4:
```suggestion
```

backend/app/models/invoice.rb
```diff
@@ -218,8 +218,7 @@ def mark_as_paid!(timestamp:, payment_id: nil)
   end
 
   def calculate_flexile_fee_cents
-    fee_cents = BASE_FLEXILE_FEE_CENTS + (total_amount_in_usd_cents * PERCENT_FLEXILE_FEE / 100)
-    [fee_cents, MAX_FLEXILE_FEE_CENTS].min.round
+    FlexileFeeCalculator.calculate_invoice_fee_cents(total_amount_in_usd_cents)
```
sharang-d, line 221: BASE_FLEXILE_FEE_CENTS is still present and used in this file. I think we should expose it from the new service and use it in the validation otherwise it will exist in two places, which defeats the purpose.

sharang-d (changes requested):
I came here from the dividends PR.

You say these methods are called in multiple places, but I only see one invocation each. It's a bit of premature refactoring, but I guess it's fine as it's easier for humans looking through the code to see all the fee formulae in one place.

I posted a few comments.

backend/app/services/flexile_fee_calculator.rb
```diff
@@ -0,0 +1,23 @@
...
+    percentage_fee_cents = (total_amount_in_usd_cents.to_d * INVOICE_PERCENTAGE.to_d / 100).round
+    fee_cents = INVOICE_BASE_FEE_CENTS + percentage_fee_cents
+    [fee_cents, INVOICE_MAX_FEE_CENTS].min
+  end
+
+  def self.calculate_dividend_fee_cents(total_amount_in_cents)
+    percentage_fee_cents = (total_amount_in_cents.to_d * DIVIDEND_PERCENTAGE.to_d / 100).round
+    fee_cents = DIVIDEND_BASE_FEE_CENTS + percentage_fee_cents
+    [fee_cents, DIVIDEND_MAX_FEE_CENTS].min
+  end
```
sharang-d, line 22: Will any test fail if I remove the `to_d`s? If not, we should add such tests. I've been bitten by precision in the past, which is why they're here :D

sharang-d (commented):
@nazargladish _a

backend/app/models/invoice.rb
```diff
@@ -29,10 +29,7 @@ class Invoice < ApplicationRecord
   DELETABLE_STATES = [RECEIVED, APPROVED]
   ALL_STATES = READ_ONLY_STATES + EDITABLE_STATES
 
-  BASE_FLEXILE_FEE_CENTS = 50
-  MAX_FLEXILE_FEE_CENTS = 15_00
-  PERCENT_FLEXILE_FEE = 1.5
-  private_constant :BASE_FLEXILE_FEE_CENTS, :MAX_FLEXILE_FEE_CENTS, :PERCENT_FLEXILE_FEE
+
```
slavingia, line 32: extra newline

Issue #281, opened 2025-05-08
Remove Docuseal


slavingia:
### What

See [Figma](https://www.figma.com/design/7dxyoTVv9danprBTZacNO6/Remove-docuseal?node-id=1-60&t=6bD8crYS4EzOZGnl-1) for the latest design thinking.

<img width="1232" height="928" alt="Image" src="https://github.com/user-attachments/assets/6146547d-dfb9-4e07-bb6f-905a1d9990cf" />

**MVP**

- [ ] Option to link to a signed doc when onboarding a user
- [ ] Remove Docuseal and use our own UX for signing, viewing, editing, and downloading documents
- [ ] Option to mark an uploaded PDF as already signed or unsigned and in need of signing (default, but use last value)
- [ ] Paste rich text in when creating a document for signing (always unsigned)
- [ ] Remove document type, rely on name of document
- [ ] Avoid any design changes that are unrelated to this functionality being added
- [ ] Upload and Create, Link can come later

All docs uploaded (PDFs) should be hosted via the Document model within the Flexile service, for posterity.

We will be getting rid of the concept of "document templates" and instead any unsigned document, can be used as a template for future documents.

**Future shipments**

- [ ] Implement 2-step modal
- [ ] Templates for exercise, grants, etc
- [ ] Add ability to share existing documents with a new person easily

<img width="619" height="841" alt="Image" src="https://github.com/user-attachments/assets/3a612235-b0ac-4ac1-8068-7b4e38eae03a" />

- [ ] Transmute document from PDF to rich text upon upload using AI
- [ ] Linking to an existing, outsourced document (on Google Docs for example)

### Why

Our current solution, using Docuseal, is ugly and complex. It also introduces a commercial dependency.

laugardie:
Made some [changes](https://www.figma.com/design/7dxyoTVv9danprBTZacNO6/Remove-docuseal?node-id=1-60&t=6bD8crYS4EzOZGnl-1) after your feedback! lmk what you think!

slavingia:
One comment to remove the complexity of templates and just having documents (similar to how we don't have contractor templates - we did with Roles but removed).

laugardie:
Great! Simplified.
– Updated the designs to support multiple worker types (same approach as we designed before, with a few small tweaks. Would be great if you could take a look.
– Refined some of the copy.
– Tweaked the table view layout as it should also show uploaded documents with no signers or status.
[...]

PR #819, opened 2025-08-10
Remove Docuseal from equity exercises

MayaRainer:
A part of #281.

This PR replaces the current docuseal flow when exercising options with an inline signing dialog.

Before
image

After
image

## Summary by CodeRabbit

* **New Features**
  * Admins can edit a company-level "Exercise notice"; users see it when exercising options. A new signing UI (SignForm) collects electronic signatures.

* **Bug Fixes / UX**
  * Option exercise flow simplified to a single-step sign-and-submit interaction with a clearer modal and reduced signing steps.

* **Chores**
  * Removed legacy environment variables tied to the previous signing integration.


frontend/app/(dashboard)/equity/grants/ExerciseModal.tsx
```diff
@@ -96,22 +93,104 @@ const ExerciseModal = ({
...
-              __companyName: company.name,
-              __name: user.legalName,
-              __email: user.email,
-              __address1: user.address.street_address,
-              __address2: `${user.address.city}, ${user.address.state} ${user.address.zip_code}`,
-              __address3: user.address.country,
-            }}
-          />
+          <>
+            <div className="prose max-h-100 overflow-auto rounded-md border p-2">
```
slavingia, line 97: I think we can have a place in settings where they can easily edit this with rich text for all the document types that would exist, since they're all tied to events happening / objects being created within the db. Would be good to save versions too per document / exercise as legal backup.

Cc @laugardie

MayaRainer: Should we provide a default, or just not allow creating grants/exercises until they've added one?

> We did a similar thing for letter of transmittal for stock buyback

I may be missing something, but it looks to me like the text there is hardcoded and not configurable, and the document isn't currently stored - in fact it looks like you can submit that form without signing at all.

slavingia: Sorry meant to say for the return of capital style dividend we did for Ludus

MayaRainer: Ah yeah makes sense! Will see what code I can share with that. Should dividend remain per-dividend (instead of per-company) but notice of exercise be per-company (instead of per-grant)?

laugardie: I can take a look at how we can handle this in settings tomorrow.

MayaRainer:
@laugardie I pushed an MVP implementation adding an exercise notice field to the equity settings:
image
The rich text provided here will be displayed during exercise and a PDF will be created from it. Currently, when a company does not have an exercise notice set, the exercise button just isn't displayed to investors, so relevant tests are failing - we probably want either some notice for admins prompting them to add this, or a fallback for the exercise notice. Feel free to make any changes, or let me know if you'd like me to make any!

laugardie:
@slavingia @MayaRainer what do you think about adding this alert here?
Screenshot 2025-08-14 at 02 40 24

I also played around with the idea for a [Templates tab](https://www.figma.com/design/egI6zF72FF4RPFIO4r0GWJ/Legal-documents-in-settings?node-id=1-773&t=o2ZlZUFFsV1mUOD3-4)  so we can group all these documents in one place. Let me know what you think.

MayaRainer:
@slavingia could you add your thoughts on ^ please? can implement it if you're fine with it :) @laugardie is there anything else or will this be good to merge once I address that?

PR #851, opened 2025-08-12
[WIP] feat(auth): Google OAuth

### LINKED ISSUE #718: Google auth
### What

Pending removal of Clerk, we should support Google login.

### Why

Faster, less friction.

lexcao:
closes #718

# Workable screen recording

https://github.com/user-attachments/assets/c56fcad9-00b7-4192-9c99-40df6a30f12a

# WIP
Tests are on the way.


lexcao:
Hi @slavingia , I am still working on this and writing test for E2E. Do you mean you are not consider this PR?

lexcao:
Hi @slavingia , I need to sync some test progress with you. The challenging part of writing E2E tests for Google OAuth is the OAuth flow. I've tried two approaches without success.

Approach 1: [source file](https://github.com/lexcao/flexile/blob/main/e2e/tests/google-oauth_not-working01.spec.ts)
- Normal Playwright operations don't work due to Google's strong bot detection.
- I referred to this [blog: Google Authentication with Playwright](https://adequatica.github.io/2023/02/06/google-authentication-with-playwright.html) for a solution.
- However, the method from the blog doesn't work on my machine; Google still blocks it.

Since bypassing bot detection is difficult, I'm considering another approach: mocking OAuth.

Approach 2: [source file](https://github.com/lexcao/flexile/blob/main/e2e/tests/google-oauth_not-working02.spec.ts)
- I learned from this [Next.js doc: Experimental Test Mode PlayWright](https://github.com/vercel/next.js/blob/v15.2.4/packages/next/src/experimental/testmode/playwright/README.md) how to mock fetch request in Next.js.
- But it's not working because `next-auth@v4` uses `openid-client@v5`, which relies on `node:http` for requests that can't be mocked by `next.onFetch`. ([source code](https://github.com/panva/openid-client/blob/v5.2.1/lib/helpers/request.js))
- I also tried node HTTP mock libraries like [nock](https://github.com/nock/nock), but they didn't work either.

Solution: In `next-auth@v5`, it uses [oauth4webapi](https://github.com/nock/nock), which supports customFetch via fetch, so upgrading NextAuth to v5 might allow us to use Next.js fetch mock for mocking the OAuth flow.

Approach 3: haven't tried yet
- Use an OAuth mock server to start a real server for testing against it.
- It's quite heavy and complex to introduce.

---

In my opinion, we can try Approach 2 in the future; mocking the OAuth flow makes focusing on our own logic (after callback logic) easier. For now, we can simply mock sessions instead. What do you think?

PR #866, opened 2025-08-12
Raise when an invoice is supposed to have equity but a grant is absent

sharang-d:
[Letting tests run before filling this out]

### DIFF
--- a/backend/app/services/invoice_equity_calculator.rb
+++ b/backend/app/services/invoice_equity_calculator.rb
@@ -12,11 +12,15 @@ class InvoiceEquityCalculator
   def calculate
     unvested_grant = company_worker.unique_unvested_equity_grant_for_year(invoice_year)
     share_price_usd = unvested_grant&.share_price_usd || company.fmv_per_share_in_usd
-    if company_worker.equity_percentage.nonzero? && share_price_usd.nil?
+    equity_percentage = company_worker.equity_percentage
+    if equity_percentage.nonzero? && share_price_usd.nil?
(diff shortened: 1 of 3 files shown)


sharang-d:
@slavingia _a what do you expect Flexile to do if a contractor is creating an invoice with 20% equity stored in the DB for them, but they don't have an equity grant for the year? We may not have handled this scenario when this feature was first written, but I'm not sure.

I feel like we should show them a prompt somewhere and disallow them from creating an invoice, but I don't remember this scenario at all, so LMK if there's some existing design I'm missing.

slavingia:
> @slavingia _a what do you expect Flexile to do if a contractor is creating an invoice with 20% equity stored in the DB for them, but they don't have an equity grant for the year? We may not have handled this scenario when this feature was first written, but I'm not sure.
> 
> I feel like we should show them a prompt somewhere and disallow them from creating an invoice, but I don't remember this scenario at all, so LMK if there's some existing design I'm missing.

It should allow them to submit, but the admin won't be able to approve (disabled button with tooltip/note) until they have created the grant for them. Ideally it's in a modal so very easy to create for them with the exact # of shares they need for the invoice autofilled!)

cc @MayaRainer

slavingia:
Cooley is running into this issue where they can submit but it always sets to 0% instead of 50%.

slavingia:
> The modal to create a grant at the time of payment approach you suggested may be a bit better as there we can tell the admin that the invoice's equity component was based on $x conversion price

I think we should go with this approach, making the equity grant inline and atomic so it's based on the current share price and also doing the document signing inline by the admin should help vs using docuseal which made it too confusing/async/easy to get out of sync.

PR #919, opened 2025-08-16
Remove json_data column and option_exercising flag checks

Anshgrover23:
- Dropped `json_flag?` method and removed :json_data column from companies
- Removed option_exercising checks in policies and presenters
- Updated equity grant and options flows to no longer rely on json_data flags
- Cleaned up frontend UI by always showing option exercising actions when applicable
- Removed jsonData from frontend DB schema and test factories

image

### DIFF
--- a/backend/db/migrate/20250816111712_remove_json_data_from_companies.rb
+++ b/backend/db/migrate/20250816111712_remove_json_data_from_companies.rb
@@ -1,0 +1,5 @@
+class RemoveJsonDataFromCompanies < ActiveRecord::Migration[8.0]
+  def change
+    remove_column :companies, :json_data, :jsonb
+  end
+end
--- a/backend/app/models/company.rb
+++ b/backend/app/models/company.rb
@@ -194,10 +194,6 @@ class Company < ApplicationRecord
     email.split("@").last
   end
 
-  def json_flag?(flag)
-    json_data&.dig("flags")&.include?(flag)
-  end
-
   def checklist_items(user)
     [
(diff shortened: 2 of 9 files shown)


Anshgrover23:
@slavingia can I get a review on this one?

backend/db/migrate/20250816111712_remove_json_data_from_companies.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveJsonDataFromCompanies < ActiveRecord::Migration[8.0]
+  def change
+    remove_column :companies, :json_data, :jsonb
```
sharang-d, line 3:
```suggestion
    remove_column :companies, :json_data, :jsonb, default: {"flags" => []}, null: false
```

To make a rollback add the correct column.

sharang-d (commented):
@slavingia perhaps this should exist as a setting in the UI at some point (Carta had it, IIRC). I used it to turn off exercising right around the time of the dividends, so the cap table didn't change. We probably don't need to do that now with the dividend creation UI in place, but in case you do for that/something else, we should add it to the UI as a company administrator setting.

@Anshgrover23 _a please resolve conflicts.

slavingia:
I'd rather not remove json_data and do one thing per PR

Anshgrover23:
@slavingia json_data is only used for one flag ```option_exercising```
thus we can remove it.

slavingia:
@laugardie can suggest something for that!

PR #922, opened 2025-08-16
Fix UI bugs

apdev2:
- Fixes bug in invoices page where emptying values from certain cells is resulting in breaking of the UI.

gif1
gif3
- Improve validation logic in `QuantityInput` for numeric and duration formats. Ensure input correctness on blur and add UI feedback when user enters incorrect values for that field. NaN's would not occur now.

gif4

- Click `x` on onboarding modals will now work.

gif2

### DIFF
--- a/frontend/app/(dashboard)/documents/FinishOnboarding.tsx
+++ b/frontend/app/(dashboard)/documents/FinishOnboarding.tsx
@@ -15,9 +15,10 @@ type OnboardingStepProps = {
   open: boolean;
   onNext: () => void;
   onBack: () => void;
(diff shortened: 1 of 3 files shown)


apdev2:
Hello! Noticed couple of issues while navigating flexile and raised this PR with fixes for the same.

apdev2:
> Let’s hide the red invalid border when the field is empty, it should only appear after the user enters something invalid.

Updated this behavior

apdev2:
> update e2e specs to cover game-breaking bugs

Added this. Having some trouble running it locally. Debugging into it.

PR #923, opened 2025-08-16
fix(auth): improve OTP flow with numeric-only validation, auto-submit, and clearer redirect states

Anshgrover23:
**Before**

* Users could type letters/symbols in OTP field.
* Validation showed a generic `"string must contain 6 characters"`.
* Button flickered active before redirect.
* No explicit "Redirecting..." feedback.

https://github.com/user-attachments/assets/16840bdc-a50c-4221-b53f-fdbc89aff2ce

**After**

* OTP field accepts numbers only.
* Specific error message shown for invalid input.
* All inputs/buttons disabled during verification and redirect.
* Clear inline status with spinner for **Verifying… → Redirecting…**.
* Auto-submit works seamlessly when all 6 digits are entered.

https://github.com/user-attachments/assets/a6b66e02-b9de-4843-b311-43109419fe87

### DIFF
--- a/frontend/app/(auth)/index.tsx
+++ b/frontend/app/(auth)/index.tsx
@@ -4,6 +4,7 @@ import { useMutation } from "@tanstack/react-query";
 import Image from "next/image";
 import { useRouter, useSearchParams } from "next/navigation";
 import { getSession, signIn } from "next-auth/react";
+import { useState } from "react";
 import { useForm } from "react-hook-form";
 import { z } from "zod";
 import { MutationStatusButton } from "@/components/MutationButton";
@@ -35,6 +36,7 @@ export function AuthPage({
 }) {
   const router = useRouter();
   const searchParams = useSearchParams();
+  const [redirectInProgress, setRedirectInProgress] = useState(false);
   const sendOtp = useMutation({
     mutationFn: async (values: { email: string }) => {
       const response = await request({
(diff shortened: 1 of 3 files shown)


Udit-takkar:
@Anshgrover23 thanks for fixing this. I was about to open a PR for this. It was really confusing before when the loading stopped before redirect

Anshgrover23:
@Udit-takkar Beat you to it 😉 — I had already fixed and pushed this, so this PR should wrap it up.

laugardie:
@Anshgrover23 
I think we can just always show “Verifying your code...” instead of showing "Verifying..." and “Redirecting...”.
Also, is it possible to display the errors without pushing the Back to email button down?

Thanks for making this change! 🙏

frontend/app/(auth)/index.tsx
```diff
@@ -144,27 +151,24 @@ export function AuthPage({
...
-                <div className="pt-6 text-center">
-                  <Button
-                    className="text-gray-600"
-                    variant="link"
-                    onClick={() => sendOtp.reset()}
-                    disabled={verifyOtp.isPending}
-                  >
-                    Back to email
-                  </Button>
+                <div className="pt-5 text-center">
```
laugardie, line 161:
```suggestion
                <div className="text-center">
```

laugardie:
@Anshgrover23 looks great! Just a small nit -> added a commit suggestion.

Anshgrover23:
@laugardie done

https://github.com/user-attachments/assets/fba9b962-4e93-41e4-a30d-c40842d32e02

slavingia:
We should remove the spacing so it feels like one input like most OTP forms

Anshgrover23:
@slavingia @laugardie Is this fine?
image

Anshgrover23:
@slavingia 
is this fine.

image

laugardie:
Something like this @Anshgrover23 
Screenshot 2025-08-19 at 15 24 04

PR #959, opened 2025-08-19
Add ParseInvoiceDocService to support use of existing invoice PDF

the-wunmi:
Partially fixes #937

Usage: This can be used on rails console using

```ruby
file = File.open("./invoice.pdf")
company = Company.last
ParseInvoiceDocService.new(file:, company:).process
```

Passing specs:

image


slavingia:
Should be entirely front-end

Issue #823, opened 2025-08-11
GitHub auth


slavingia:
After Google auth

slavingia:
If you want, that's fine I think.

PR #963, opened 2025-08-19
Feat: add GitHub Auth

andev18:
Resolves https://github.com/antiwork/flexile/issues/823

### Result
https://github.com/user-attachments/assets/7d8efbcf-08d7-401d-93fa-c478b6a7e711

### E2e tests
Screenshot 2025-08-19 at 10 31 29 PM
Screenshot 2025-08-19 at 10 01 33 PM
Screenshot 2025-08-19 at 10 06 22 PM


kanhayaKy:
> Looks good overall. Let me kick off the tests and get the right env variables in on our side.

@pallavi I think we should hold on to this as I have raised concerns on this being duplicated from my branch 
I have written this to @slavingia

slavingia:
Yes this has someone else’s commits in it. Best to let them take this to completion.

slavingia:
Since the code is pattern matching what has already been merged into main, this can be taken through completion and the bounty will be split proportionally ($1k for this, $1.5k for the other).

PR #966, opened 2025-08-19
Support drag-and-drop of existing invoice PDF

CyMule:
Closes #937

@slavingia you mentioned that this should potentially be a front end only change, but I do not think that is the case unless I am misunderstanding how you want this to work. This needs the backend changes in order to securely make the OpenAI API calls.

## Tests Pass Locally
image

## Video Demo 1 (clicking import from pdf button on top)

This video shows
1. Only PDF's uploads are allowed
2. If the PDF is not an invoice, a dismissable error message is shown
3. If a PDF invoice is uploaded, the information is automatically filled in

https://github.com/user-attachments/assets/1f270905-ec67-4623-bb67-539857a0bef4

## Video Demo 2 (showing drag and drop)

https://github.com/user-attachments/assets/3c11a355-58dc-4c08-92c7-f89a08ef9cb2

## Feedback Questions

Do you have thoughts on the UI?
Should we have the import button + drag and drop?
Should we add a cancel button on the processing pop up?
Should I use add VCR cassette to this PR for the OpenAI requests?


CyMule:
I converted the changes to typescript / NextJS only.

## Tests still pass

Screenshot 2025-08-19 at 3 06 44 PM

## Video demo
https://github.com/user-attachments/assets/7047937f-37c9-4aa2-b708-572bfd263c8b

frontend/hooks/usePdfExtraction.ts
```diff
@@ -0,0 +1,163 @@
...
+  const fileInputRef = useRef<HTMLInputElement>(null);
+
+  const showError = (message: string) => {
+    // eslint-disable-next-line no-console
+    console.error("PDF extraction error:", message);
+    setError(message);
+  };
+
+  const processFile = async (file: File) => {
+    if (file.type !== "application/pdf") return;
```
CyMule, line 48: Should we allow other file types or focus on just pdfs?

PR #968, opened 2025-08-20
feat: allow to drag and drop files when creating a support ticket

VladSez:
Drag and drop files

https://github.com/user-attachments/assets/1b3a0c9e-1f1d-415b-b2c6-c9cd9ce35034

In case file is not supported nothing happens:

https://github.com/user-attachments/assets/87c5d613-6a53-42db-875f-f8e8a2e411f4

- Added react-dropzone to enable drag-and-drop file uploads.
- Implemented file acceptance criteria for images, PDFs, DOC, DOCX, and TXT files.
- Enhanced UI to provide visual feedback during file attachment process.
- Updated ConversationsList component to manage file attachments more effectively.

@slavingia Should we also add frontend validations to prevent uploading, for example, files larger than 5 MB each?

### DIFF
--- a/frontend/app/(dashboard)/support/ConversationsList.tsx
+++ b/frontend/app/(dashboard)/support/ConversationsList.tsx
@@ -1,8 +1,9 @@
 "use client";
 
 import { useConversations, useCreateConversation } from "@helperai/react";
-import { CircleCheck, Paperclip, Plus, SendIcon, X } from "lucide-react";
+import { CircleCheck, Paperclip, Plus, SendIcon, Upload, X } from "lucide-react";
 import React, { useRef, useState } from "react";
+import { useDropzone } from "react-dropzone";
 import { helperTools } from "@/app/(dashboard)/support/tools";
 import { DashboardHeader } from "@/components/DashboardHeader";
 import { MutationStatusButton } from "@/components/MutationButton";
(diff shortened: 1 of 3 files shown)


frontend/app/(dashboard)/support/ConversationsList.tsx
```diff
@@ -150,27 +169,40 @@ export const ConversationsList = ({ onSelectConversation }: ConversationsListPro
...
+                </div>
               </div>
             </div>
 
+            {isDragActive ? (
+              <div className="absolute inset-0 z-10 mb-0 flex items-center justify-center rounded-lg border-2 border-dashed border-blue-400 bg-blue-50/90">
+                <div className="text-center">
+                  <Upload className="mx-auto h-12 w-12 text-blue-400" />
+                  <p className="mt-2 text-sm font-medium text-blue-600">Drop files here to attach</p>
+                  <p className="text-xs text-blue-500">Supports images, PDF, DOC, DOCX, and TXT files</p>
```
slavingia, line 201: not sure if needed, the person can assume all work?

slavingia:
adding an end to end spec would be great

frontend/app/(dashboard)/support/ConversationsList.tsx
VladSez: I thought this message would work better from a UX perspective. I used the accepted file formats listed here
image

btw should we add some file size validations on the FE? i.e. allow files only up to 5mb or 10mb each, wdyt?

VladSez:
@slavingia sorry why this was closed? =)

PR #970, opened 2025-08-20
feat(auth): add password auth

pallavi:
## What

| Sign up | Log in |
|--|--|
| <video src="https://github.com/user-attachments/assets/006e2e20-8384-4ff0-8312-b16cf2549f4e" controls muted playsinline style="max-width:320px;width:100%;height:auto;border-radius:12px;"></video> | <video src="https://github.com/user-attachments/assets/37221097-faec-416e-917f-65f2cc232358" controls muted playsinline style="max-width:320px;width:100%;height:auto;border-radius:12px;"></video> |

## Follow ups in future PRs
- Ability to set and update password in Settings page. Needs design. I think I know where it would go, but here are open questions:
  - Do we want the user to re-type their old password?
  - Do we want the user to type the new password twice to check that the two values match?
  - If we do both of the above, would we want to do this in a modal?
- Ability to reset password. Do we need to prioritize this? A user could always gain access to their account through OTP verification, and then set their password from the Settings page...but this could be seen as an inconvenient workaround.


backend/app/controllers/internal/signup_controller.rb
```diff
@@ -30,14 +30,24 @@ def send_otp
...
+      temp_user = find_temp_user(email)
+      return unless temp_user
 
-    return unless check_otp_rate_limit(temp_user)
-    return unless verify_user_otp(temp_user, otp_code)
+      return unless check_otp_rate_limit(temp_user)
+      return unless verify_user_otp(temp_user, otp_code)
+    else
+      temp_user = User.new(email: email, password: password)
+      temp_user.save!(validate: false) # Skip validations for temp user
```
kanhayaKy, line 49: We created a user with the provided email and password but did not verify if the email belongs to them, this could be a security issue as I can sign up with any random email ?

kanhayaKy (commented):
@pallavi One small issue in the new user sign up flow, we are not verifying if the provided email belongs to the user trying to create the account, for sign up we should be taking the password and also the OTP to ensure this. What are your thoughts on this @slavingia ?

backend/app/controllers/internal/signup_controller.rb
kanhayaKy: Here's how I'd imagine the signup controller to change 

```diff
  def verify_and_create
    email = params[:email]
    otp_code = params[:otp_code]
+   password = params[:password]

    return unless validate_signup_params(email, otp_code)

    temp_user = find_temp_user(email)
    return unless temp_user

    return unless check_otp_rate_limit(temp_user)
    return unless verify_user_otp(temp_user, otp_code)

    # Check again if user was created in the meantime
    existing_user = User.find_by(email: email)
    if existing_user && existing_user.id != temp_user.id
      temp_user.destroy
      return render json: { error: "An account with this email already exists. Please log in instead." }, status: :conflict
    end

+    if !password.blank?
+      temp_user.password = password
+    end

    # Complete user creation
    result = complete_user_signup(temp_user)

    if result[:success]
      success_response_with_jwt(result[:user], :created)
    else
      render json: { error: result[:error_message] }, status: :unprocessable_entity
    end
  end
```

laugardie:
@pallavi
- can you add a break after password?
Screenshot 2025-08-21 at 06 47 35

I’ll also review the follow-up questions and propose a design.

PR #977, opened 2025-08-21
Fix failing specs

the-wunmi:
Introduced a change on https://github.com/antiwork/flexile/commit/1a1280e3d96cd4f96a67486065eddafdc5a8fbf7 that's now causing specs to fail, this addresses the issue.

image


the-wunmi:
@MayaRainer I didn’t revalidate the specs after making the letter of transmittal column not nullable as requested, and now the specs are failing. This fixes the issue. Sorry about the oversight.

Issue #868, opened 2025-08-12
Incorrect contract ended date


chaitanyya:
### What

If I end a contract today for a person (testing in PT time), the list shows the date that is -1 day from today for me.

eg. Contract ended on 12 Aug, the dates listed in the people list is 11 Aug.

### Why

This is inconsistency in how we store dates and how we show it to the user. Do we always want to show localized time or do we want to preserve some timezone? (we should list it on the status)

The tests definitely fail on my machine not sure how are they passing in CI if they are at all.

**Ending Contract:**
<img width="2034" height="1142" alt="Image" src="https://github.com/user-attachments/assets/f38bbf76-9470-48ad-a025-8e8f36e93460" />

**People List:**
<img width="2034" height="1142" alt="Image" src="https://github.com/user-attachments/assets/52bc59f0-8d10-41d8-8e95-4ac0d42aaad9" />

PR #983, opened 2025-08-21
fix: incorrect contract ended date

andev18:
Resolves https://github.com/antiwork/flexile/issues/868

### Root cause
The `contractor.endedAt` is a server date but was not properly formatted as a server date using `formatServerDate` before getting displayed. So the displayed end date was shifted backwards when the timezone is `UTC−08:00`/`UTC−07:00` because the front-end tries to convert an UTC date to a date of such timezone.

This doesn't happen on CI because on CI the browser date is also UTC.
`
### Before
The contract end date was set to Aug 21.
Screenshot 2025-08-21 at 9 14 08 AM

### After
Screenshot 2025-08-21 at 9 13 37 AM

### E2e tests still passing
Screenshot 2025-08-21 at 9 19 18 AM


slavingia:
Missing specs

PR #985, opened 2025-08-21
Use more workers for Playwright on CI

MayaRainer:


MayaRainer (commented):
@slavingia _a up to you whether you'd like to merge this; it'll be a bit more expensive overall as the bigger runners cost twice as much per minute, and the speedup seems to be about 30%

Issue #947, opened 2025-08-19
All-in-one Docker


slavingia:

chethanuk:
@slavingia Does Antiwork have Discord?

PR #1007, opened 2025-08-22
Adding support docker for env to enable more oss contrib - #947

chethanuk:
Closes: #947

Important commands:

- Start Docker env:
Screenshot 2025-08-22 at 4 27 58 AM
- Multi containers to check docker logs easily 
Screenshot 2025-08-22 at 5 18 22 AM

I recommend dev's setting up mise and use - but if antiwork community don't want mise [I can make changes to just use bash scripts] - Check 1. https://mise.jdx.dev/ and [2](https://spr.com/simplifying-multi-language-development-with-mise/) - I recommend mise - it's pretty cool

WIP: Still 2 or 3 errors need to be fixed for WSL or linux users


docker-compose.dev.yml
```diff
@@ -0,0 +1,242 @@
...
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    networks:
+      - flexile-network
+
+  # MinIO S3-compatible storage
+  minio:
```
chethanuk, line 55: Also let me know if we dont want minio - many users or devs from some countries wont have AWS access or may be we can create docker-compose.dev.yml without minio or S3 localstack and one more docker-compose.dev.minio.yml with minio

PR #1016, opened 2025-08-23
ci: improve test workflow for playwright job

sm17p:
## Description

This PR aims to decrease test workflow runtime for the playwright job

### How?

1. For playwright browser, avoid downloading the full browser and just download the shell required for headless mode run
2. Following https://github.com/actions/cache/issues/92, refactors caching logic to manually save cache before running the test suite. The manual control allows for saving resources in cases where tests don't pass, which makes them usable on the next run

### AI Disclosure
1. deepwiki used to verify caching directories for playwright and puppeteer

### Checklist
- [x] Self-reviewed


sm17p:
> Since the build takes only ~30s, do you think it’s worth maintaining a relatively complex workflow and caching logic?

|[First run](https://github.com/antiwork/flexile/actions/runs/17441694792/job/49526054851)|[Second Run](https://github.com/antiwork/flexile/actions/runs/17443810953/job/49533094322)|
|---|---|
|R1|R2|

It's the change in caching logic which reduced the build time from ~1m to ~40s, as the project grows this is likely to increase. 

About maintenance, this is a one time effort and I think this workflow won't need a re-touch untill `turbopack` is stable in nextjs. 

In unison with stuff in #1045 and #1069, overall it does impact DX positively by having a faster feedback loop. Having local maching running the whole test suite is a big blocker to multi-tasking

PR #1029, opened 2025-08-23
fix: inviting existing users as lawyers

sm17p:
### Why?

> Found out about this while working working on phase 2 of https://github.com/antiwork/flexile/pull/828, which will cover improvements to the user selection box in add member modal. I sliced out https://github.com/antiwork/flexile/pull/1029 from that work as a temporary measure to address it, otherwise I could open with phase 2 altogether. Let me know!

https://github.com/antiwork/flexile/issues/1028#issuecomment-3217586628

### Description

Invite flow was broken for existing users, lawyers couldn't be invited but admins could

1. Fixed it by removing the blocking path
2. Tried to reduce the number of queries while at it
3. Specs to cover the existing and new case

### Before

https://github.com/user-attachments/assets/6b5ead3f-b8f9-4e53-950e-99022b5fbe32

### After

https://github.com/user-attachments/assets/f747b473-bd06-4e48-b7e3-ccdf7ab55dd8

#### Backend

backend

#### Frontend

frontend

### AI Disclosure

1. Deepwiki used to understand flexile schema
2. Pair programming session with claude to cover the edge cases

### Chekllist

- [x] Self reviewed


backend/app/services/invite_lawyer.rb
```diff
@@ -3,30 +3,36 @@
...
-      error_object = if company_lawyer.errors.any?
-        company_lawyer
-      else
-        user
-      end
-      { success: false, field: error_object.errors.first.attribute, error_message: error_object.errors.first.full_message }
+    ActiveRecord::Base.transaction do
+      user = User.find_or_create_by!(email: @email)
+      company_lawyer = user.company_lawyers.create!(company: @company)
+      user.invite!(@current_user) { |u| u.skip_invitation = true }
```
sm17p, line 16: `skip_invitation` This was related to clerk, right? Can we remove it?

sm17p (commented):
Reviewed! can be taken up

Anshgrover23:
@sm17p already fixed permanently in #1030 
No need for your temporary fix.

PR #1039, opened 2025-08-25
Fix decimal input in bank verification microdeposit amounts

### LINKED ISSUE #1010: Can't verify bank because input doesn't take decimal point
The two small deposits have a decimal point, but I can't enter the decimal in the input fields. The fields seem to accept only integer numbers.

Image

CyMule:
Closes #1010

## Before
before-decimals

## After
image

## Demo Video
[demo.webm](https://github.com/user-attachments/assets/d0129c5c-5850-4ae8-8f2d-e81f98e4fe92)

## AI Disclosure

AI was used for an initial draft of this PR, and I had to manually go step by step to get the test to work correctly. 

## Self review

I reviewed the code in this PR (with one comment) and locally by verifying the changes through my e2e test.

## Tests

The updated test passes without issue
image

See [this comment](https://github.com/antiwork/flexile/pull/1039#discussion_r2298414250) on changes I would need to make to get the other payment test to pass for all contributors

I did run all tests, and there are quite a few failing. This is a minimal change so I don't believe any tests are failing as a result of these changes
image

### DIFF
--- a/e2e/tests/settings/administrator/payment-details.spec.ts
+++ b/e2e/tests/settings/administrator/payment-details.spec.ts
@@ -71,31 +71,69 @@ test.describe("Company administrator settings - payment details", () => {
     await login(page, adminUser);
     await page.getByRole("link", { name: "Settings" }).click();
     await page.getByRole("link", { name: "Billing" }).click();
-    await page.getByRole("button", { name: "Link your bank account" }).click();
-
-    const stripeFrame = page.frameLocator("[src^='https://js.stripe.com/v3/elements-inner-payment']");
-    await stripeFrame.getByRole("button", { name: /Enter bank details manually/u }).click();
-
-    const bankFrame = page.frameLocator("[src^='https://js.stripe.com/v3/linked-accounts-inner']");
-    await bankFrame.getByLabel("Routing number").fill("110000000");
-    await bankFrame.getByTestId("manualEntry-accountNumber-input").fill("000123456789");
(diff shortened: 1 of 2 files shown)


e2e/tests/settings/administrator/payment-details.spec.ts
```diff
@@ -63,39 +63,85 @@ test.describe("Company administrator settings - payment details", () => {
...
+    await page.getByRole("button", { name: "Verify bank account" }).click();
+    await expect(page.getByRole("dialog")).toBeVisible();
+
+    // Stripe determines at runtime whether microdeposits use two "amounts" or a "descriptor_code" (6-digit code).
+    // The Payment Element does not allow forcing a specific type, so the test handles both.
+    const codeInput = page.getByLabel("6-digit code");
+    if (await codeInput.isVisible().catch(() => false)) {
+      await codeInput.fill("SM11AA");
+      await expect(codeInput).toHaveValue("SM11AA");
+    } else {
```
CyMule, line 126: > I was working on updating the first test and it looks like this may not be deterministic. Stripe doesn't seem to always return the amounts verification. You can see in the video demo that it did, but when I was working on the second test I keep getting the descriptor code (6 digit code) verification.
> 
> I could update this test to handle both and/or work on the mocking as well.

Ideally we would have a cleaner separation here. At the moment, it seems we need to handle the `amounts` and `descriptor_code` case.

I think we can just ignore what stripe returns in dev in do either case, but I would need to set up some mocking to be sure.

slavingia:
@CyMule can you look into the failing lints and fix?

PR #1041, opened 2025-08-26
fixed bin/dev and port on next config.ts

kaushal07wick:
bin/dev automatically pulls from vercel env variables which automatically overrides the `.env` file.
fixed it to just echo.

and 

was getting un-configured host after the updates from #1017 
the port should be 3000 not 3001.

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -3,7 +3,7 @@
 set -e
 
 if [ -f ".vercel/project.json" ]; then
-    pnpx vercel env pull .env
+    echo "Skipping \`vercel env pull\` to avoid overriding .env"
 elif [ ! -f ".env" ]; then
     echo ".env file not found. Please run bin/setup first."
     exit 1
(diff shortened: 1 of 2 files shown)


kaushal07wick:
@ershad

bin/dev
```diff
@@ -3,7 +3,7 @@
 set -e
 
 if [ -f ".vercel/project.json" ]; then
-    pnpx vercel env pull .env
+    echo "Skipping \`vercel env pull\` to avoid overriding .env"
```
ershad, line 6: It would pull only if `.vercel/project.json` exists, right? Antiwork internal team uses Vercel to manage envs, so we need this here.

frontend/next.config.ts
```diff
@@ -26,7 +26,7 @@ const nextConfig: NextConfig = {
   },
 };
 if (process.env.NODE_ENV === "development") {
-  nextConfig.images.remotePatterns.push({ protocol: "http", hostname: "localhost", port: "3001" });
+  nextConfig.images.remotePatterns.push({ protocol: "http", hostname: "localhost", port: "3000" });
```
ershad, line 29: Thanks for this change.

We can merge the PR if you could revert the changes in `bin/dev`.

PR #1083, opened 2025-09-04
Remove the extra mx-4 from the alert in Workspace settings

laugardie:
Screenshot 2025-09-04 at 07 39 23


laugardie:
@slavingia _a

PR #1084, opened 2025-09-04
Disabled inputs style

laugardie:
Before
Screenshot 2025-09-04 at 07 53 01

After
Screenshot 2025-09-04 at 07 52 18


laugardie:
@slavingia _a

PR #1094, opened 2025-09-05
Add document templates

MayaRainer:
A part of #281. Allows creating templates from within our UI that are used when creating new documents.

image

image


MayaRainer (commented):
@laugardie _a could you review please? added a few more related comments in the figma file too!

laugardie:
@MayaRainer _a I reviewed your comments in Figma.
I agree that making the edit template live in a modal would work well after ditching the history. Here’s a [reference](https://www.figma.com/design/egI6zF72FF4RPFIO4r0GWJ/Legal-documents-in-settings?node-id=38-1733&t=0pni6Hr2EZk4MD3Q-4).

> What do you mean by "We could show all documents related to the account"?

I meant only showing the templates tied to the features enabled. For example, if equity isn’t enabled, it wouldn’t make sense to display equity-related templates.

I suggest changing:
Consulting contract – (Used for) New contractors → Consulting agreement – (Used for) Contracts

laugardie:
@MayaRainer looking good! Just one small tweak, could we make the modal a bit wider to match the [design](https://www.figma.com/design/egI6zF72FF4RPFIO4r0GWJ/Legal-documents-in-settings?node-id=38-1733&t=0pni6Hr2EZk4MD3Q-4)?

MayaRainer:
Done! Will merge this, but if there's anything else I can tweak it in a follow-up!

PR #1104, opened 2025-09-06
Add share document functionality

MayaRainer:
A part of #281.

### DIFF
--- a/backend/app/controllers/internal/companies/documents_controller.rb
+++ b/backend/app/controllers/internal/companies/documents_controller.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class Internal::Companies::DocumentsController < Internal::Companies::BaseController
+  def create
+    authorize Document
+    document = Current.company.documents.build(**document_params, year: Time.current.year)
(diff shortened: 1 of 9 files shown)


MayaRainer (commented):
@laugardie _a could you review this one as well please? :)

laugardie:
@MayaRainer _a
I think we should follow the same pattern as in Invoices, right-click anywhere in the row should open the contextual menu with possible actions, download, share...We could also simplify the ui by moving all secondary actions into the menu and keeping only primary actions like "Sign" directly in the row.

https://github.com/user-attachments/assets/93356889-d944-49c4-9651-be79d4cfc8ea

Screenshot 2025-09-08 at 11 40 39


Would you mind removing the "Document" label and reducing the spacing to gap-4 in the Share document modal?

Screenshot 2025-09-08 at 12 03 26

> Are these two versions of the same thing? Which one is the current? 

[Link to comment](https://www.figma.com/design/7dxyoTVv9danprBTZacNO6?node-id=1-60#1413531476)

MayaRainer:
@laugardie _a
> Do you know if there's a reason why consulting agreements aren’t downloadable?

I think that's just because the ones generated in seeds don't have attachments. Shouldn't be an issue in production.

Addressed the other comments, lmk if this is good now!

PR #1108, opened 2025-09-07
fix: overlapping row count text on invoices page

Salman-in:
Part of #911 

- [x] Fixes selected row count display text on 3+ digits

## Before 
Screenshot from 2025-09-07 10-56-37

## After
Screenshot from 2025-09-07 10-56-55

## Tests
Simple CSS change (no breaking changes)

## AI Disclosure
No usage of AI


laugardie:
@slavingia _a looks good to me!

PR #1120, opened 2025-09-09
[WIP] Handle equity checks during invoice approval rather than creation

### LINKED ISSUE #882: Ensure equity splits are not wiped when submitted
- Instead, admin should need to create the grant before approving will go through

MayaRainer:
Fixes #882.

### DIFF
--- a/backend/app/models/invoice.rb
+++ b/backend/app/models/invoice.rb
@@ -79,6 +79,7 @@ class Invoice < ApplicationRecord
   validate :total_must_be_a_sum_of_cash_and_equity
   validate :min_equity_less_than_max_equity
   validate :deleted_invoices_cannot_have_active_only_status
+  validate :worker_must_have_sufficient_unvested_shares
 
   scope :pending, -> { alive.where(status: COMPANY_PENDING_STATES) }
   scope :processing, -> { where(status: PROCESSING) }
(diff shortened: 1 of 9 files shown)


MayaRainer:
> I think we should also consider removing batch approvals in the front-end at least.

That would make things simpler at least for this purpose, yes. If you're confident, I'll make that change @slavingia.

> Recipient info is already prefilled and disabled. We could also prefilled the number of options needed.

We can prefill the user, but not anything else as things like "relationship to company" can't be determined from the data we have per user. Probably don't want to prefill the number of options needed either as in most cases you would probably want to create more shares than just enough to cover a specific invoice.

> For batch approval. If invoices have enough shares individually but exceed the grant in batch, approve the ones that fit.

To expand on my previous comment, the problem with this, and with the async nature of invoice payments (and thus vesting) in general, is that we don't know at approval time how many shares are currently pending to be vested from pending invoice payments. Specifically, it happens [here](https://github.com/antiwork/flexile/blob/0619ae3a95aef9efda8bf11dd890969e09aae531/backend/app/models/invoice.rb#L192) after the invoice has already been paid. I think in order to solve this properly, we would have to do a bigger refactor where we associate invoices with equity grants when we trigger the payment (right now we [try to select an applicable grant during vesting and raise an error if it fails](https://github.com/antiwork/flexile/blob/0619ae3a95aef9efda8bf11dd890969e09aae531/backend/app/sidekiq/vest_stock_options_job.rb#L15), but this is never communicated to the user as it happens during a Wise webhook), and add a field per grant indicating how many unvested shares are currently "reserved" for pending invoice payments. Anything else is likely to lead to (or rather, continue causing) intermittent shadow payment failures as we end up without an applicable grant at a point where the invoice should already be considered "done". cc @slavingia as this would increase the scope quite a bit, and cc @sharang-d in case you have anything else to add!

> When issuing payments, as contractors need to approve and later the admin need to pay. I think we could follow the same approach as mentioned above.

As far as I can tell, the admin does not have to manually trigger a payment for this; it's triggered when the user accepts it. Don't think the approach above would work then, right?

slavingia:
Commented; should we update an e2e test w this full flow?

MayaRainer:
> Commented

Sorry I don't see it, where did you comment @slavingia?

If we do decide to go with the big refactor then yes it should absolutely be properly E2E tested to ensure reserving the shares works properly.

PR #1138, opened 2025-09-11
fix leave workspace success handling

kanhayaKy:
AI Disclosure
No AI was used

This PR handles the leave workspace success scenario. Instead of redirecting the user to various pages, it now logs them out, allowing them to start a fresh session. This approach prevents the unintended creation of a new company when a user isn't associated with one, reducing confusion for users who don't wish to continue. Those who do want to continue can simply log back in and will be redirected to an existing or new company.

Demo

https://github.com/user-attachments/assets/0f0d89ba-63db-48e1-b0ba-cc0f178fd06c

E2E
Screenshot 2025-09-12 at 12 17 22 AM

Part of #911 and https://github.com/antiwork/flexile/issues/1132


MayaRainer (commented):
@kanhayaKy _a two questions/comments!

MayaRainer:
Thanks!

PR #1159, opened 2025-09-14
Add dialog stack component

MayaRainer:
A part of #281.


MayaRainer (commented):
@laugardie _a Could you have a look at this and make the necessary design changes please? I had to make quite a few changes to the component both to support the reverse order and to fix the generally rather low code quality, but it seems to work well now!

laugardie:
@MayaRainer _a pushed some changes. Could you remove the ‘Contract’ label from the second modal and add the ‘x’ close button like we have in the single modal?

https://github.com/user-attachments/assets/38e9bd32-4063-49ed-8ddb-39e403fab07e

slavingia:
Awesome!

frontend/components/ui/form.tsx
```diff
@@ -73,7 +73,7 @@ function FormItem({ className, ...props }: React.ComponentProps<"div">) {
 
   return (
     <FormItemContext.Provider value={value}>
-      <div data-slot="form-item" className={cn("grid grid-rows-[auto_auto_1fr] gap-2", className)} {...props} />
+      <div data-slot="form-item" className={cn("grid gap-2", className)} {...props} />
```
laugardie, line 76: `grid-rows-[auto_auto_1fr]` was adding an extra spacing to all form inputs. Do you think is ok to update it to this? I've checked some forms and they seem to be ok.

laugardie (commented):
@MayaRainer _a I've pushed a few changes to improve the ui/ux.
There are still a few things pending, if you could help with that.
- The x in the modal doesn't seem to work.
- In the New grant modal the continue button when disabled shouldn't allowed to navigate to the next modal until the necessary info is complete.
- I'm not sure how the cropped focus in New grant modal can be properly fix.
- Could you also fix the spacing in add a contract and equity contract in the footer? as they are inside the form I don't know what is the best approach to fix this.

Screenshot 2025-09-18 at 13 08 42
Screenshot 2025-09-18 at 13 09 00

MayaRainer:
@laugardie _a Made the changes, lmk if there's anything else!

laugardie:
@MayaRainer _a

There’s still some extra space below the button because of the grid-rows-[auto_minmax(0,1fr)_auto]. Since the buttons in both examples are inside the form, do you think the best fix is just to remove the auto?
Screenshot 2025-09-22 at 10 15 50

In the Equity Grant flow, after selecting the recipient and the relationship to the company auto-fills, the Continue button stays disabled but I think it shouldn't.

In the Vesting details modal and the Equity contract modal, the Back button doesn’t seem to be working.

We should probably make header and footer always visible and middle content scrollable... I tried before but couldn’t get it working.

PR #1167, opened 2025-09-15
fix(documents): add download button to mobile view

nyomanjyotisa:
### Explanation of Change
Fixes an issue where the download button was missing from the documents list on mobile devices. This prevented users from downloading their signed documents directly from mobile.

### Screenshots/Videos
Before
image

After
image

### AI Disclosure
No AI tools used


slavingia:
> @nyomanjyotisa FYI, this is fixed in #1073

Much better to fix incrementally like this than larger PRs.

PR #1177, opened 2025-09-15
Fix csrf token issue caused by #1017

kanhayaKy:
AI Disclosure
No AI was used

Description
As part of #1017 https and flexile.dev was removed, this caused the csrf token not being saved by the browser due to the secure and domain options provided. This  breaks certain flows that depend on cookies. For example switching the company does not work. 

Before

https://github.com/user-attachments/assets/023d8f82-cfb5-4018-b4c1-47658d6f1045

After

https://github.com/user-attachments/assets/2e5720ea-e391-4cad-a5ee-244809e3cc17

Tests
No changes in tests as tests don't depend on this.


ershad:
Thanks for the PR! Focusing on Gumroad this week, will test/review next week.

PR #1180, opened 2025-09-15
fix: settings profile update payload 

abhix4:
Ref:- https://github.com/antiwork/flexile/issues/911

## Description 

- previously, the settings form was sending user data inside a nested settings object
- the backend controller only permits params (email, preferred_name).
- causing, updates to profile information (like email and preferred name) were not being saved.

### Before

[Screencast from 2025-09-16 00-24-28.webm](https://github.com/user-attachments/assets/a703a9c8-863c-4c56-b130-c9d31e0ff6de)

### After

[Screencast from 2025-09-16 00-25-01.webm](https://github.com/user-attachments/assets/96112462-6057-4638-a6c7-6e792cc3baa3)

#### AI Disclosure:-

### DIFF
--- a/frontend/app/settings/page.tsx
+++ b/frontend/app/settings/page.tsx
@@ -37,6 +37,7 @@ export default function SettingsPage() {
 }
 
 const DetailsSection = () => {
+  const queryClient = useQueryClient();
   const user = useCurrentUser();
   const form = useForm({
     defaultValues: {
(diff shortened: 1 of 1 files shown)


abhix4:
@slavingia please review.

abhix4:
> Please resolve merge conflicts

I think this was already fixed in #1224  but I completely understand how easy it is to miss when there are many open contributions. This is actually the third time I’ve run into a similar situation, so I just wanted to kindly suggest: maybe if we try reviewing PRs starting from the bottom (oldest first), it could help avoid duplicate efforts. Totally understand if it’s not always possible, just thought I’d share the idea 🙏.
@slavingia

PR #1183, opened 2025-09-16
fix(flaky tests): stabilise tax settings test by ordering userComplianceInfos by createdAt

romitg2:
### Root Cause:
- The test queried userComplianceInfos without an explicit order, so the “latest” record wasn’t guaranteed—leading to flaky assertions.

### Solution:
Added orderBy: [desc(userComplianceInfos.createdAt)] to always fetch the newest record first. so in assertions we always get latest record at `0th` index.

### AI Disclosure:
- no AI used

### CI Failure Link:
https://github.com/antiwork/flexile/actions/runs/17728605580/job/50374743417?pr=1173#step:12:433
Screenshot 2025-09-16 at 9 36 13 AM

### After:
Screenshot 2025-09-16 at 4 11 47 PM

### DIFF
--- a/e2e/tests/settings/tax.spec.ts
+++ b/e2e/tests/settings/tax.spec.ts
@@ -9,9 +9,9 @@ import { usersFactory } from "@test/factories/users";
 import { fillDatePicker, selectComboboxOption } from "@test/helpers";
 import { login } from "@test/helpers/auth";
 import { expect, test } from "@test/index";
-import { eq } from "drizzle-orm";
+import { desc, eq } from "drizzle-orm";
 import { BusinessType, TaxClassification } from "@/db/enums";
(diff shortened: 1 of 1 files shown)


e2e/tests/settings/tax.spec.ts
```diff
@@ -152,10 +152,13 @@ test.describe("Tax settings", () => {
         .findFirst({
           where: eq(users.id, user.id),
           with: {
-            userComplianceInfos: true,
+            userComplianceInfos: {
+              orderBy: [desc(userComplianceInfos.createdAt)],
```
romitg2, line 156: so that during assertions(below) we get latest info at 0th index. (before order wasn't guaranteed so tests were flaky)

PR #1203, opened 2025-09-18
Invoice stale data

romitg2:
### Bug / Problem:
- stale data was displayed on the invoice edit page after a user successfully updated an invoice and then navigated back to the edit page.

### Solution:
- invalidate the cached data after update mutation succeeds.

### AI Disclosure:
- some help to write test

# Before: 
https://github.com/user-attachments/assets/97d747a8-543a-4e04-ae92-26a0aa1ef906

# After:
https://github.com/user-attachments/assets/ab3e4c7a-6917-48c0-a3f6-7515e25e8e7b


romitg2:
@slavingia RFR!

Issue #1115, opened 2025-09-08
Fixes for company investors with 0 shares


slavingia:
- Don't loop through these when issuing dividends
- Specs
- deleted_at?

sharang-d:
@slavingia I wonder if there should be two flags: one for when a user wants to leave the company/workspace, and another for when they should not have access to data because they sold all their holdings or similar.

They're two separate states. For example, when an investor has "left the company workspace" they have done so only on the app, but they're still an active investor from the company's POV. They would still be considered when we want to show how many investors would be eligible for a dividend (as an example). Whereas if they're no longer an investor, that'd mean we simply ignore them everywhere, like a deleted record, but preserve it for historical purposes.


WDYT?

sharang-d:
I'm a little unsure if we should even let an active investor (meaning one that holds any kind of security) leave a company's workspace. For example, it'd be odd to send a dividend to such an investor's bank account but them seeing no details about it on account of having left the company's workspace. 

Leaving the workspace probably makes more sense for contractors over investors.

slavingia:
Probably leaving should just remove it as a company contractor. It's mostly to just remove it from the nav switcher.

PR #1222, opened 2025-09-22
Don't delete `CompanyInvestor` when leaving a workspace

sharang-d:
### What

Don't delete a `CompanyInvestor` record from the system when a user is leaving a company's workspace.

### Why

A company investor record is referenced by multiple tables in the database. Also, even if it were a soft-delete, we shouldn't allow the investor themself to do it. It should be administrator/system-initiated, so we only do it after the investor holds no securities of the company.

This was discovered because the dividends page was broken for a customer. It seems an investor deleted their own record in the system using this feature.

### Notes

AI Usage: Cursor Tab
Ref: https://github.com/antiwork/flexile/issues/1115#issuecomment-3319889166, https://anti--work.slack.com/archives/C068HHZ64M9/p1757889317135169?thread_ts=1757333132.357889&cid=C068HHZ64M9

### DIFF
--- a/backend/app/policies/user_policy.rb
+++ b/backend/app/policies/user_policy.rb
@@ -10,6 +10,6 @@ class UserPolicy < ApplicationPolicy
   end
 
   def leave_company?
-    (company_worker.present? || company_investor.present? || company_lawyer.present?) && company_administrator.blank?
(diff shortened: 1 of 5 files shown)


sharang-d:
@slavingia _a this will stop the dividends page from breaking like it did for the customer recently.

PR #1234, opened 2025-09-23
fix: templates page on small screens

abhix4:
Ref:- https://github.com/antiwork/flexile/issues/911

## Description 

- fix the templates page on small screens

### Before

image

### After

image

#### AI Disclosure:-


laugardie:
@abhix4 _a added an example of the table [here](https://www.figma.com/design/L4SUjxa7JOfnhq9d0HCCAO/Mobile-navigation?node-id=1127-2465&t=vVFJTGwjKTm1m7Ww-4) for reference. Key points to note:
- The table should be full width.
- If there’s no template, show the button only.
- On mobile, make sure removing the header doesn’t leave empty space. Currently the header isn’t visible but its element still takes up space.
- Use black text for both the title and the “used for” label.
- On mobile, the top line in the table can be removed.

PR #1241, opened 2025-09-24
Fix tax page bug

kanhayaKy:
AI Disclosure
No AI was used

Part of #911 

What?
The tax page does not throw a validation error for some required fields on clicking `save changes` hence the save fails silently. 

Before

https://github.com/user-attachments/assets/7264cbc0-7ed8-4bc2-ac81-dd8d2416a3ec

After

https://github.com/user-attachments/assets/3d2b16ce-325e-41a4-a81d-9d693f3faf9a


frontend/app/settings/tax/page.tsx
```diff
@@ -185,6 +185,9 @@ export default function TaxPage() {
 
     if (values.country_code === "US" && !/(^\d{5}|\d{9}|\d{5}[- ]\d{4})$/u.test(values.zip_code))
       return form.setError("zip_code", { message: "Please add a valid ZIP code (5 or 9 digits)." });
+
+    if (countrySubdivisions.length > 0 && values.state.length === 0)
```
MayaRainer, line 189: Should we cover this in the E2E test too? otherwise looks good! @kanhayaKy _a

PR #1244, opened 2025-09-24
fix: make invoice with expense item editable 

sm17p:
### Ref
- #911

Invoice editing expects expense `id` to be a string but currently it's passed as a number causing a crash on the page.

### Before

https://github.com/user-attachments/assets/16bbecbb-74e2-4791-93ad-733d726b529a

### After

Addressed the crash by

1. Proxying `id` usage with string based `external_id` for expense records throughout the codebase 

https://github.com/user-attachments/assets/9d62a6a2-11f2-40d6-9f30-00931df992fe

### AI Usage
1. Claude for edits


backend/app/presenters/invoice_presenter.rb
```diff
@@ -96,7 +96,7 @@ def expenses
       invoice_expenses.reject { _1.marked_for_destruction? }
                       .map do |expense|
         {
-          id: expense.id,
+          id: expense.external_id,
```
sm17p, line 99: Pass `external_id` instead of raw id in REST endpoints

backend/db/migrate/20250923081928_add_external_id_to_invoice_expenses.rb
```diff
@@ -0,0 +1,21 @@
...
+  def change
+    add_column :invoice_expenses, :external_id, :string
+
+    up_only do
+      InvoiceExpense.where(external_id: nil).find_each do |expense|
+        InvoiceExpense::ExternalIdGenerator::ID_MAX_RETRY.times do
+            external_id = Nanoid.generate(size: InvoiceExpense::ExternalIdGenerator::ID_LENGTH,
+                                          alphabet: InvoiceExpense::ExternalIdGenerator::ID_ALPHABET)
+            unless InvoiceExpense.where(external_id:).exists?
+              expense.update_columns(external_id:)
```
sm17p, line 11: Update existing entries with unique `external_id` before `indexing` adding and `not null` constraint

sm17p: Don't know the table size so sharing alternative option

```ruby
class AddExternalIdToInvoiceExpenses < ActiveRecord::Migration[8.0]
  def change
    add_column :invoice_expenses, :external_id, :string
    
    up_only do
      batch_size = 5000
      existing_ids = Set.new(InvoiceExpense.pluck(:external_id).compact)
      total_rows = InvoiceExpense.where(external_id: nil).count
      max_iterations = (total_rows / batch_size) + 1
      
      max_iterations.times do
        batch = InvoiceExpense.where(external_id: nil).limit(batch_size).to_a
        break if batch.empty?
        
        generated_ids = Set.new
        while generated_ids.size < batch.size
          external_id = Nanoid.generate(
            size: InvoiceExpense::ExternalIdGenerator::ID_LENGTH,
            alphabet: InvoiceExpense::ExternalIdGenerator::ID_ALPHABET
          )
          generated_ids.add(external_id) unless existing_ids.include?(external_id)
        end
        
        update_records = batch.zip(generated_ids.to_a).map do |expense, external_id|
          {
            id: expense.id,
            external_id: external_id
          }
        end
        
        InvoiceExpense.upsert_all(
          update_records,
          unique_by: :id,
          update_only: [:external_id]
        )
        
        existing_ids.merge(generated_ids)
        
        puts "Updated #{batch.size} invoice expenses with external IDs"
      end
      
      remaining_count = InvoiceExpense.where(external_id: nil).count
      if remaining_count > 0
        raise "Migration incomplete: #{remaining_count} records still missing external_id"
      end
    end
    
    add_index :invoice_expenses, :external_id, unique: true
    change_column_null :invoice_expenses, :external_id, false
  end
end
```

slavingia:
missing e2e test to prevent future regressions

PR #1245, opened 2025-09-25
Remove document name field in favour of document type

### LINKED ISSUE #281: Remove Docuseal

MayaRainer:
Closes #281.

### DIFF
--- a/backend/app/models/document.rb
+++ b/backend/app/models/document.rb
@@ -12,52 +12,39 @@ class Document < ApplicationRecord
 
   has_many_attached :attachments
 
-  CONSULTING_CONTRACT_NAME = "Consulting agreement"
-
-  FORM_1099_DIV = "1099-DIV" # Dividends and Distributions
-  FORM_1099_NEC = "1099-NEC" # Nonemployee Compensation
-  FORM_W_9 = "W-9"
-  FORM_1042_S = "1042-S"
-  FORM_W_8BEN = "W-8BEN"
-  FORM_W_8BEN_E = "W-8BEN-E"
-  SUPPORTED_TAX_INFORMATION_NAMES = [
-    FORM_W_9,
-    FORM_W_8BEN,
-    FORM_W_8BEN_E,
-  ].freeze
-  SUPPORTED_IRS_TAX_FORM_NAMES = [
-    FORM_1099_NEC,
-    FORM_1099_DIV,
-    FORM_1042_S,
-  ].freeze
-  ALL_SUPPORTED_TAX_FORM_NAMES = SUPPORTED_TAX_INFORMATION_NAMES + SUPPORTED_IRS_TAX_FORM_NAMES
+  SUPPORTED_TAX_INFORMATION_TYPES = %i[form_w_9 form_w_8ben form_w_8ben_e].freeze
+  TAX_FORM_TYPES = %i[form_1099_nec form_1099_div form_1042_s form_w_9 form_w_8ben form_w_8ben_e].freeze
 
   validates_associated :signatures
-  validates :name, presence: true
   validates :document_type, presence: true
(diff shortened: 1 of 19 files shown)


laugardie:
looks good to me as well.

backend/app/models/document.rb
```diff
@@ -12,64 +12,82 @@ class Document < ApplicationRecord
...
+    when "form_1042s"
+      "1042-S"
+    when "form_w9"
+      "W-9"
+    when "form_w8ben"
+      "W-8BEN"
+    when "form_w8bene"
+      "W-8BEN-E"
+    end
+  end
```
sharang-d, line 84: As a contractor/investor on Flexile, I'd prefer to see these names on my documents:

```
- Equity Plan Contract 2023
- Equity Plan Contract 2024
- GUM-99 Share Certificate
```

vs 

```
- Equity Plan Contract
- Equity Plan Contract
- Share Certificate
```

Unless I'm mistaken, with this changeset we'll use the second set of names, which isn't an improvement in my eyes. I am aware that there's a date in the UI per document, so having the year in the filename may be redundant, but it helps for tax forms and such (this isn't the case ATM FWIW so consider it as an improvement), where the issue date would be a year later, and for share certificates, it's a must to differentiate between multiple.

(I see this point has been made on the original issue already, but I'll keep it here since this **is** my feedback)

Secondly, with this naming method, a change here would mean all my documents use a new name, which I don't like, as I consider these to be "immutable" in a sense, and changing the name does not signal that.

MayaRainer:
@slavingia feel free to merge if you don't think the points outlined by Sharang are an issue!

backend/app/models/document.rb
slavingia: Can we resolve this by having a def name method that just cocantenates the equity plan contract with the year for example, the share certificate with the ID?

slavingia:
Responded to that comment.

backend/app/models/document.rb
sharang-d: Yes, that'd be better!

MayaRainer: We could do that, but not retroactively as currently documents aren't associated with share holdings in any way, so the old ones would still just be called "Share Certificate". Is that fine?

slavingia: that's fine, we can manually migrate if it's an issue I think

MayaRainer:
@slavingia _ a feel free to merge unless there's anything else!

slavingia:
some merge conflicts

MayaRainer:
@slavingia _a done

PR #1236, opened 2025-09-24
fix: show missing status for people in mobile view

kr-sushil:
### Description:

The People mobile view doesn't show the status text. Previously, a status icon was displayed, but since those have been removed, the corresponding text should now be shown.

Part of #911 

### Before / After:

 <table>
 <tr><th>Before</th><th>After</th></tr>
 <tr>
 <td> 
Screenshot 2025-09-24 at 10 18 46 AM
 </td>
 <td>
Screenshot 2025-09-24 at 10 16 13 AM
 </td>
 </tr>
 </table>

For reference: status text shown in desktop view
 <table>
 <tr>
 <td> 
Screenshot 2025-09-24 at 10 29 23 AM
 </td>
 </tr>
 </table>

### Test Suite / Related tests:
 <table>
 <tr>
 <td> 
Screenshot 2025-09-24 at 10 21 26 AM
 </td>
 </tr>
 </table>

> [!Note]
> AI Disclosure: No AI was used to generate any of this code.
>Self-review: All changes are intuitive, so no additional comments are needed.

PR #1253, opened 2025-09-25
fix: status label styling in tables view

kr-sushil:
### Description:

Update status text styling from multiple colour to black/white colour (based on theme) in  table views. This is based on discussion in this [thread](https://github.com/antiwork/flexile/pull/1236#issuecomment-3327716044).

PR covers:
- Updates table (Admin view)
- Documents table
- Billing table

Part of #911 

### Before / After:

- Added in self review coments

> [!Note]
> AI Disclosure: No AI was used to generate any of this code.


laugardie:
@slavingia _a looks good to me! I still think badges would help, but this works for me.

PR #1260, opened 2025-09-25
fix billing spec

kanhayaKy:
Part of #1132 

AI Disclosure:
No AI used

Description:
The stripecustomerId should be passed as null for tests when creating company, otherwise the company factory creates a random id 

Failing test : e2e/tests/settings/administrator/billing.spec.ts
Failing CI Run : https://github.com/antiwork/flexile/actions/runs/18017207360/job/51265139296

E2E
image


ershad:
We can assume that customer (`cus_M2QFeoOFttyzTx`) exists on Stripe when we run tests from CI. But even then, it doesn't explain why the test is **flaky** and how adding `stripeCustomerId: null` when creating company in e2e tests could concretely fix the issue in `billing.spec.ts`.

Closing the PR for now, but please feel free to reopen / send another PR if you have more info.

PR #1261, opened 2025-09-25
payment spec fix due to playwright strict mode violation

VarunArora14:
## Problem
The "shows 'Pay again' button for failed payments" test was failing with a Playwright strict mode violation(default setting):

image

This occurred because there were multiple `<tbody>` elements on the page, and Playwright's strict mode requires locators to resolve to exactly one element.

Allowing first occurence to be visible is consistently passing the test.

image

fixes https://github.com/antiwork/flexile/issues/1132

AI Disclosure:
No AI tools used


VarunArora14:
@kanhayaKy probably yes, felt similar for using first() method. I think PRs resolving single tbody might be able to fix this better. Closing this for now.

PR #1256, opened 2025-09-25
fix: correct user compliance info assertions in tax settings tests

VarunArora14:
## Problem
The test "allows confirming tax information" was failing because it had incorrect assertions about which compliance info records should be active vs deleted.

When a user updates their tax information the system then - 
1. Creates a new `userComplianceInfo` record with updated data
2. Marks the old record as deleted (`deletedAt` is set)
3. Orders results by `createdAt DESC` (newest first)

Screenshot 2025-09-26 012549

The test was incorrectly expecting:
- `userComplianceInfos[0]` (newest) to be deleted 
- `userComplianceInfos[1]` (oldest) to be active 

### Fix

Corrected the assertions to properly expect:
- `userComplianceInfos[0]` (newest) to be active (`deletedAt = null`) 
- `userComplianceInfos[0]` to have `taxInformationConfirmedAt` set 
- `userComplianceInfos[1]` (oldest) to be deleted (`deletedAt != null`)

## Changes
- Fixed test assertions in `e2e/tests/settings/tax.spec.ts`
- No functional code changes, only test corrections

Result Screenshot - 

image

Ref https://github.com/antiwork/flexile/issues/1132

AI Disclosure:
No AI was used

PR #1266, opened 2025-09-25
Fix(flaky test): tax spec was failing because createdAt timestamp inconsistency

romitg2:
ref #1132 

### What PR Does?
 resolve timestamp inconsistency in tax compliance test by using consistent JavaScript dates

  ## Problem
- The test was flaky due to timestamp inconsistencies between different record creation methods
- i logged the complianceInfo and saw that `createdAt` has different dates (which supposed to be few miliseconds of difference)
Screenshot 2025-09-26 at 4 10 17 AM

### Solution:
 Made factory timestamps consistent**: Updated `userComplianceInfosFactory.create()` to explicitly use `createdAt: new Date()` instead of relying on database `defaultNow()`

### AI Disclosure:
- used claude-4 for brainstorming

### DIFF
--- a/e2e/factories/userComplianceInfos.ts
+++ b/e2e/factories/userComplianceInfos.ts
@@ -30,6 +30,7 @@ export const userComplianceInfosFactory = {
       .values({
         userId: user.id,
         taxId: "000-00-0000",
+        createdAt: new Date(),
         ...nonTaxComplianceAttributeDefaultValues,
         ...overrides,
       })
--- a/e2e/tests/settings/tax.spec.ts
+++ b/e2e/tests/settings/tax.spec.ts
@@ -164,10 +164,10 @@ test.describe("Tax settings", () => {
 
       expect(updatedUser.userComplianceInfos).toHaveLength(2);
 
-      expect(updatedUser.userComplianceInfos[0]?.deletedAt).not.toBeNull();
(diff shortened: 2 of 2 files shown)


VarunArora14:
The test is not flaky and seems to be failing consistently. Refer here for explanation - https://github.com/antiwork/flexile/pull/1256#issuecomment-3341152494

I have created a PR before solving this issue as a heads up - https://github.com/antiwork/flexile/pull/1256

cc @slavingia

ershad:
> i logged the complianceInfo and saw that createdAt has different dates (which supposed to be few seconds of difference)
> Updated userComplianceInfosFactory.create() to explicitly use createdAt: new Date() instead of relying on database defaultNow()

Do we know how `defaultNow()` can result in different dates?

romitg2:
> I see. `5:30` difference is surprising, but how can two sequential `defaultNow()` call can lead to it? Are we we able to reproduce this difference consistently / frequently?

no idea why this is happening. consistently happens

ershad:
I see. It is also unlikely to be the cause for flakiness in CI because system timezone is UTC. I will close this PR in favor of #1256, but please feel free to reopen / send new PR if you find any new info.

slavingia:
We cannot merge code unless we understand why things are happening in the first place

PR #1272, opened 2025-09-26
Support past investors as a user type in Flexile

sharang-d:
Refer: #1115

### DIFF
--- a/backend/db/migrate/20250926102109_add_deactivated_at_to_company_investors.rb
+++ b/backend/db/migrate/20250926102109_add_deactivated_at_to_company_investors.rb
@@ -1,0 +1,5 @@
+class AddDeactivatedAtToCompanyInvestors < ActiveRecord::Migration[8.0]
+  def change
+    add_column :company_investors, :deactivated_at, :datetime
+  end
+end
--- a/backend/db/schema.rb
+++ b/backend/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_09_05_123957) do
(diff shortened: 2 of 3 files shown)


backend/spec/services/publish_company_update_spec.rb
```diff
@@ -25,14 +25,14 @@
         end.to change { CompanyUpdateEmailJob.jobs.size }.by(2)
 
         expect(CompanyUpdateEmailJob).to have_enqueued_sidekiq_job(company_update.id, active_company_worker.user_id)
-        expect(CompanyUpdateEmailJob).to have_enqueued_sidekiq_job(company_update.id, company_investor.user_id)
+        expect(CompanyUpdateEmailJob).to have_enqueued_sidekiq_job(company_update.id, active_company_investor.user_id)
       end
 
-      it "doesn't enqueue email jobs for inactive contractors or lawyers" do
+      it "doesn't enqueue email jobs for inactive contractors or investors" do
```
sharang-d, line 31: Spec was incorrect as lawyers have nothing to do with this class.

frontend/trpc/routes/tenderOffers/index.ts
```diff
@@ -64,7 +67,10 @@ export const tenderOffersRouter = createRouter({
   }),
 
   get: companyProcedure.input(z.object({ id: z.string() })).query(async ({ ctx, input }) => {
-    if (!ctx.company.equityEnabled || (!ctx.companyAdministrator && !ctx.companyInvestor))
+    if (
+      !ctx.company.equityEnabled ||
+      (!ctx.companyAdministrator && !(ctx.companyInvestor && !ctx.companyInvestor.deactivatedAt))
+    )
```
sharang-d, line 73: These conditions will be easier on the eyes if/when the endpoints are moved to REST APIs, as we can then use policies.

PR #1282, opened 2025-09-26
fix: restrict the EIN number length

abhix4:
REF //  https://github.com/antiwork/flexile/issues/911

## Description 

Users were able to type more than 9 digits in the EIN field. Although validation prevented invalid EINs from being saved, this created a poor user experience.
- Added input length restriction.

### Before
image

### After

Not able to type more
image

#### AI Disclosure:-


romitg2:
@abhix4 I already have PR #1163 for this

PR #1295, opened 2025-09-28
fix: simplify slider interaction for equity percentage selection in one-off-payments

VarunArora14:
Ref: https://github.com/antiwork/flexile/issues/1132

fixes flaky test: https://github.com/antiwork/flexile/actions/runs/18074387433/job/51428747827

The E2E test "with an allowed equity percentage range" was flaky and failing intermittently. The test was expecting exact equity percentage values of 25-75% but was getting slightly different values (26-74%) due to imprecise mouse drag interactions with the range slider

Replaced unreliable mouse drag interactions with precise keyboard navigation for setting slider values

Verify test using: `pnpm test:e2e --grep "with an allowed equity percentage range"`

image

AI discloure:
Github copilot used


VarunArora14:
This is similar resolution like -  https://github.com/antiwork/flexile/pull/1143/

I verified other specs that this is the last implementation of using such code with page mouse movements.

@slavingia  please review this

PR #1300, opened 2025-09-28
fix: improve user display in roles page with better layout

Muhammad-Sheik-Nauman:
Part of #911

### Description
---
Fixed text overflow on smaller and medium screens by allowing names to wrap or truncate properly, improving readability and layout consistency.

### Before
---
before

### After
---
after

### AI Disclosure
---
No AI usage.


slavingia:
No, sorry

PR #1333, opened 2025-09-30
fix(failing tests): One-off payments › invoice list visibility

sm17p:
### Ref
- #1132 

### What's Happening?

The current custom table matcher has multiple failure points 

1.  Stuck asserting against stale skeleton loader

https://github.com/user-attachments/assets/278c9d0d-65c8-49a2-93af-7374b5c299d1

2. No row found to assert against

Screenshot 2025-09-30 at 3 07 27 PM

### The fix
1. Address the techdebt marker by removing `matchers.ts` in favour of ~~web first assertions~~ custom `getByTableRow` locator and `tableRow` selector engine
2. ~~Add 'role' attribute to `table` components which enables locating them in a stable fashion~~ Add scope `col` for `TableHead` component which makes it selectable via getByRole("columnheader") - https://github.com/antiwork/flexile/pull/1333#pullrequestreview-3290789241

### Test Addressed - [Link to failed run from main](https://github.com/antiwork/flexile/actions/runs/18118644153/job/51559439204)

1. `e2e/tests/company/invoices/one-off-payments.spec.ts:293:5 › One-off payments › invoice list visibility › does not show one-off payments in the admin invoice list while they're not accepted by the payee` 
2. `tests/company/invoices/one-off-payments.spec.ts > One-off payments › invoice list visibility › shows 'Pay again' button for failed payments` 

### Specs Passing Local

one-off-payments-specs-passing

### Specs Passing in CI
1. Run 1 - https://github.com/antiwork/flexile/actions/runs/18124108413/job/51575259636
2. Run 2 - https://github.com/antiwork/flexile/actions/runs/18125228683/job/51578824522

### AI Usage
- Claude for adding aria roles


e2e/helpers/matchers.ts
```diff
@@ -1,33 +0,0 @@
...
-
-    if (matchesAll) {
-      return row;
-    }
-  }
-
-  return null;
-};
-
-// TODO (techdebt) clean this up
```
sm17p, line 28: Cleaned this up in favour of web first assertion, seemed more reasonable than fixing this

e2e/tests/company/invoices/one-off-payments.spec.ts
```diff
@@ -314,12 +313,14 @@ test.describe("One-off payments", () => {
...
 
-      const invoiceRow = await findRequiredTableRow(page, {
-        "Invoice ID": "O-0001",
-        Amount: "$123.45",
-      });
+      await page
+        .getByRole("row")
+        .nth(1)
+        .filter({ has: page.getByRole("cell", { name: "O-0001" }) })
+        .filter({ has: page.getByRole("cell", { name: "$123.45" }) })
```
sm17p, line 320: This chaining works because Playwright's `Locator` methods return new `Locator` instances that can be further refined. It can be chained to filter multiple times.

frontend/components/ui/table.tsx
```diff
@@ -42,6 +51,7 @@ function TableRow({ className, ...props }: React.ComponentProps<"tr">) {
         className,
       )}
       {...props}
+      role="row"
```
sm17p, line 54: Adding aria roles to encourage usage of `page
        .getByRole("row")` assertions. 

Two advantages of this over css selectors

1. Role-based selectors encourage writing tests that focus on what elements do rather than how they are implemented
2. These are most of the time resilient to changes in the DOM unless there's a major re-write

sm17p: > The reason we use tbody tr is to exclude header rows - they could also be written as locator("tbody").getByRole("row"), but the former is simpler

Makes sense. How do you feel about `page.getByRole("row").nth(1)`? I used it cause I find it more intuitive

e2e/helpers/matchers.ts
```diff
@@ -1,33 +1,31 @@
...
+    if (trimmedText) {
+      map[trimmedText] = index;
     }
-  }
+    return map;
+  }, {});
 
-  return null;
-};
+  const columnIndexes = Object.keys(columnValues).reduce<Record<string, number>>((result, label) => {
```
sm17p, line 17: Maps user provided columns to column number

e2e/helpers/matchers.ts
```diff
@@ -1,33 +1,31 @@
...
+    const index = allHeaderTextToIndex[label];
+    assert(index !== undefined, `Could not find column header "${label}" in table.`);
```
sm17p, line 19: fails early when header cell is not present

fails early when header cell is not present

MayaRainer (commented):
@sm17p _a Looks like that's a Playwright bug where `th[scope]` doesn't default to `col` - see [here](https://stackoverflow.com/questions/77944459/why-does-playwright-think-the-aria-role-for-a-th-tag-is-cell). We can set `scope="col"` on `th` which should fix this!

e2e/helpers/matchers.ts
```diff
@@ -0,0 +1,31 @@
+import { expect, type Locator, type Page } from "@playwright/test";
+import { assert } from "@/utils/assert";
+
+export const findTableRow = async (page: Page, columnValues: Record<string, string>): Promise<Locator> => {
```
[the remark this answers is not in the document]
sm17p: Still working on it synchronous is a bit complicated

e2e/helpers/matchers.ts
```diff
@@ -1,31 +1,29 @@
...
-      rows.filter({
-        has: page.locator(`td:nth-child(${(columnIndexes[columnLabel] ?? 0) + 1})`, { hasText: expectedValue }),
-      }),
-    page.locator("tbody tr"),
-  );
+  return Object.entries(columnValues).reduce((rows, [columnLabel, expectedValue]) => {
+    const columnIndex = allHeaderTextToIndex.get(columnLabel);
+    return rows.filter({
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- await expect(headerRowByColumnNames).toBeVisible(); guarantees this never happen
+      has: page.locator(`td:nth-child(${columnIndex! + 1})`, { hasText: expectedValue }),
```
sm17p, line 26: await expect(headerRowByColumnNames).toBeVisible(); guarantees columnIndex is always present

e2e/index.ts
```diff
@@ -1,16 +1,41 @@
...
+import { createTableRowEngine, serializeColumnValues } from "@test/selectors/tableRowEngine";
 import { test as baseTest } from "next/experimental/testmode/playwright.js";
 import type { CreateEmailOptions } from "resend";
 import { parseHTML } from "zeed-dom";
 import { assertDefined } from "@/utils/assert";
 
 export * from "@playwright/test";
 
+type ExtendedPage = Page & {
+  getByTableRowCustom: (columnValues: Record<string, string | RegExp>) => Locator;
```
sm17p, line 11: Our custom locator, usage

```js
page.getByTableRowCustom({ "Invoice ID": "O-0001", Amount: "$123.45"})
```

e2e/index.ts
```diff
@@ -1,16 +1,41 @@
...
+    async ({ playwright }, use) => {
+      await playwright.selectors.register("tableRow", createTableRowEngine);
+      await use();
+    },
+    { scope: "worker", auto: true },
+  ],
+  page: async ({ page }, use) => {
+    const extendedPage: ExtendedPage = Object.assign(page, {
+      getByTableRowCustom: (columnValues: Record<string, string | RegExp>) =>
+        page.locator(`tableRow=${serializeColumnValues(columnValues)}`),
```
sm17p, line 35: This is how you'd use the extended selector interface. For our use case it's a partial table row entry

```js
 {"Invoice ID": "O-0001", Amount: "$123.45"}
```

e2e/tests/company/equity/investors.spec.ts
```diff
@@ -111,12 +111,13 @@ test.describe("Investors", () => {
 
     await login(page, adminUser, "/equity/investors");
 
-    await expect(page.getByText("Major Investor")).toBeVisible();
-    await expect(page.locator("tbody")).toContainText("15.00%");
-    await expect(page.locator("tbody")).toContainText("300,000");
-
-    await expect(page.getByRole("cell", { name: "Outstanding ownership" })).toBeVisible();
```
sm17p, line 118: `scope="col"` broke this as it would be 'columnheader' now, took the chance to test out the custom locator

e2e/index.ts
```diff
@@ -24,12 +49,6 @@ export const test = baseTest.extend<{
     });
     await use(emails);
   },
-  setup: [
-    async ({}, use) => {
```
sm17p, line 28: I think this fixes an issue which I've seen, I'll tag when I find it

e2e/selectors/tableRowEngine.ts
```diff
@@ -0,0 +1,91 @@
...
+      }, new Map<string, number>());
+
+      const rowsToCheck = targetTable.querySelectorAll("tbody tr");
+      const matchingRows = Array.from(rowsToCheck).filter((tableRow) => {
+        const rowCells = tableRow.querySelectorAll("td");
+        return Object.entries(searchCriteria).every(([columnName, expectedValue]) => {
+          const columnIndex = columnNameToIndexMap.get(columnName);
+          if (columnIndex === undefined) return false;
+
+          const cell = rowCells[columnIndex];
```
sm17p, line 58: Check for the search string in the exact table cell (x,y)

e2e/selectors/tableRowEngine.ts
```diff
@@ -0,0 +1,91 @@
...
+            const compiledRegex = new RegExp(regexPattern, regexFlags);
+            return compiledRegex.test(actualCellValue);
+          }
+          return actualCellValue.includes(expectedValue);
+        });
+      });
+
+      return options.findAll ? matchingRows : (matchingRows[0] ?? null);
+    } catch {
+      return options.findAll ? [] : null;
```
sm17p, line 79: Playwright requires selector engines to not throw

PR #1346, opened 2025-09-30
Fix buttons responsiveness and styles

laugardie:


laugardie:
@slavingia _a can you take a look?

PR #1348, opened 2025-09-30
fix: billing settings test - ensure company data sync when name is null

HiAbhishekh:
Part of https://github.com/antiwork/flexile/issues/1132

AI Disclosure: Used for only for syntax, logic and code review Manually

### Problem
The Playwright test "billing settings gated until company name is set" was failing intermittently in CI (job #51591463961) due to a data synchronization issue between the database and frontend.

### Locally it was falling

Screenshot 2025-09-30 at 8 35 26 PM

### Root Cause :
The test was creating a company with name: null directly using createPreOnboarding()
However, the frontend was not receiving the updated company data immediately
This caused the alert message to either not appear or show incorrect content
The flakiness occurred because the frontend sometimes cached the company data or didn't sync with the database changes

### How Addressed This Issue:
Database Synchronization: Instead of creating a company with null name directly, we create it with a valid name first, then update it to null in the database
Frontend Refresh: Added await page.reload() to force the frontend to fetch the latest company data from the database
Reliable Assertion: The test now consistently finds the alert element because the frontend has the correct company data

### After: Now it is Passing
Screenshot 2025-09-30 at 8 43 06 PM

### Testing: Verified locally that the test now passes consistently!

Fixes: #1132

### DIFF
--- a/e2e/tests/settings/administrator/billing.spec.ts
+++ b/e2e/tests/settings/administrator/billing.spec.ts
@@ -1,31 +1,28 @@
+import { db } from "@test/db";
 import { companiesFactory } from "@test/factories/companies";
+import { companyAdministratorsFactory } from "@test/factories/companyAdministrators";
+import { usersFactory } from "@test/factories/users";
 import { login } from "@test/helpers/auth";
 import { expect, test } from "@test/index";
+import { eq } from "drizzle-orm";
+import { companies } from "@/db/schema";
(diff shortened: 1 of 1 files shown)


HiAbhishekh:
hii @slavingia can you please review this ?

MayaRainer:
Closing this as not the/a correct solution.

PR #1355, opened 2025-09-30
fix: invoice equity calculation test - use flexible text matching

HiAbhishekh:
Part of : #1132 

AI Disclosure: Used for only for syntax, logic and code review Manually

## Problem
CI test failure: "considers the invoice year when calculating equity"
- Error: `expect(locator).toBeVisible()` - `getByText('Total services$6,000')` not found
- Test timeout of 30000ms exceeded

Flakey/Before:https://github.com/antiwork/flexile/actions/runs/18129189994/job/51591463961

### Root Cause Analysis

The test was failing because of **exact text matching** that was too strict for CI environments:

Screenshot 2025-09-30 at 11 33 06 PM

After:
Screenshot 2025-09-30 at 11 35 58 PM

### Solution
 Use flexible text matching with `{ exact: false }` option
 
###  Testing:
Test passes locally Consistently

### DIFF
--- a/e2e/tests/company/invoices/create.spec.ts
+++ b/e2e/tests/company/invoices/create.spec.ts
@@ -72,9 +72,9 @@ test.describe("invoice creation", () => {
     await page.getByLabel("Hours / Qty").fill("100:00");
     await page.getByPlaceholder("Description").fill("I worked on invoices");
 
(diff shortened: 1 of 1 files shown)


HiAbhishekh:
hii @slavingia can you please Review?

PR #1363, opened 2025-10-01
 Fix: for company investors with 0 shares 

JagjeevanAK:
Ref: #1115 

Now it correctly work's as @slavingia described in ISSUE 1115 Also I have tried to update the test's 

## AI Disclosure
No AI is used to fix this issue


slavingia:
bad english

PR #1375, opened 2025-10-01
fix(flaky test): flaky contractor invite link spec

romitg2:
ref: #1132

### AI Disclosure:
- cursor used for debugging

### failing CI link:
https://github.com/antiwork/flexile/actions/runs/18168791709/job/51718019423?pr=1367#step:12:284
Screenshot 2025-10-01 at 11 47 04 PM

# Problem:
- contractor.list fetch was pending in background, and we open invite link modal
- contractor.list fetch completes and page re-render, cause invite link modal to close before we copied the link
- now we're unable to find copy button, and test fails.

# Solution:
- wait for contractor.list to complete before opening invite link modal to avoid race condition.

## local run:
Screenshot 2025-10-02 at 12 00 55 AM


slavingia:
> cursor used for debugging

that's not enough info mate.

PR #1394, opened 2025-10-06
remove acceptance by contractors on one-off payments

Anshgrover23:
@slavingia These were the changes which @Kavitha0809 forgot to copy from My PR.


slavingia:
Split this into one incremental change each as our policy

